PR #3, opened 2025-03-27
[CI] Update container file, run more tests in GitHub CI

pearzt:
Updates our GitHub CI to LLVM 18 and extends testing

### DIFF
--- a/.github/workflows/mcg-ci.yml
+++ b/.github/workflows/mcg-ci.yml
@@ -1,6 +1,11 @@
 name: MetaCG CI
 
-on: [push]
+on: 
+  push:
+  pull_request: 
+    types:
+      - opened
+      - synchronized
 
 jobs:
   build:
--- a/.github/workflows/mcg-ci.yml
+++ b/.github/workflows/mcg-ci.yml
@@ -38,3 +38,32 @@ jobs:
         with: 
           image: metacg-devel:latest
           run: /opt/metacg/metacg/build/pgis/test/unit/pgistests
+      - name: Run target collector tests 
+        uses: addnab/docker-run-action@v3
+        with: 
+          image: metacg-devel:latest
+          run: |
+            cd /opt/metacg/metacg/graph/test/integration/TargetCollector
+            ./TestRunner.sh
+      - name: Run call graph merge tests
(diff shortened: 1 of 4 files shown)


graph/test/integration/TargetCollector/TestRunner.sh
```diff
@@ -64,7 +64,7 @@ if [ $resultOfTest -ne 0 ]; then
...
 fi
 
-rm ./src/foo.ipcg
-rm ./src/main.ipcg
-rm ./src/wholeProgramCG.ipcg
-echo "{}" > ./src/wholeProgramCG.ipcg
+rm ./src/foo.cpp${CI_CONCURRENT_ID}.ipcg
+rm ./src/main.cpp${CI_CONCURRENT_ID}.ipcg
+rm ./src/wholeProgramCG-${CI_CONCURRENT_ID}.ipcg
+rm -rf build-${CI_CONCURRENT_ID}
```
pearzt, line 70: @TimHeldmann Could have another look at the clean-up here? I'm not sure what the `echo {} ...`  was intended to do in the first place.

TimHeldmann: If I remember correctly, cgmerge requires a valid JSON formatted file to be present as a merging target.
The `echo {} ...` command resets the file to be an empty valid JSON formatted file again.
I am not sure if we need this to run in a container, but I think it is necessary if you want to run the test manually multiple times.

pearzt: Tim and I agreed that the `echo` is no longer necessary.

sebastiankreutzer (approved):
LGTM

.github/workflows/mcg-ci.yml
```diff
@@ -1,6 +1,11 @@
 name: MetaCG CI
 
-on: [push]
+on: 
+  push:
+  pull_request: 
+    types:
+      - opened
+      - synchronized
```
jplehr, line 8: I think we can leave out these two and rely on the default.

jplehr:
I'll merge this anyway, so we see if that works as we intend it to.

PR #6, opened 2025-03-31
Correctly install generated custom MD header

sebastiankreutzer:
We generate a custom header file `CustomMD.h` that includes all of the headers added to `metadata/custom`. 
Previously, this header was not copied to the install directory. 
This patch correctly installs the generated header and excludes the `CustomMD.h.in` template file from the installation.

### DIFF
--- a/.github/workflows/mcg-ci.yml
+++ b/.github/workflows/mcg-ci.yml
@@ -17,6 +17,11 @@ jobs:
       run: cmake -B build -S .
     - name: build
       run: cmake --build build --parallel
+    - name: install
+      run: |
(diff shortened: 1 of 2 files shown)


jplehr:
Do we have a test for this problem?

sebastiankreutzer:
> Do we have a test for this problem?

No, but we can add one. Don't know much about the Github CI, but I'll try it.

sebastiankreutzer:
I've added an install step to the github CI. I seems to run fine.

jplehr (approved):
LGTM

PR #7, opened 2025-04-01
Add top-level 'metacg' folder to header install path

sebastiankreutzer:
This inserts a top-level `metacg` directory to the installed include tree.


sebastiankreutzer:
The two fixups fixed formatting and fixed the include path for the CMake target. Should be complete now.

pearzt:
@sebastiankreutzer Is this PR complete or a draft state?

PR #8, opened 2025-04-02
Add missing metadata

sebastiankreutzer:
This patch adds two new metadata types: `FilePropertiesMD` (`fileProperties` in JSON) and `InlineMD` (`inlineInfo` in JSON).
These are emitted by CGCollector but were previously not supported in the graph lib.


sebastiankreutzer:
inb4 "where are the tests": We do currently test any metadata explicitly. Definitely something we should add at some point.

TimHeldmann:
`InlineMD` is a nice addition.
As it was implemented in the cgcollector which does not use the graph lib but came after the metadata refactor patch it was never moved to the graph lib. 
Thank you.

The reason `FileMD` was not modeled as part of the graph lib is that we include an origin field as a node's core-attribute since the metadata refactor patch. 
We convert `FileMD` data to the origin attribute on read and on write as part of the V2<->V3 conversion functions.
`FileMD` was used for the example extension plugin in the cgc2 implementation.

What use-case do you have, that requires `FileMD` ?

graph/include/metadata/InlineMD.h
```diff
@@ -0,0 +1,75 @@
...
+
+ private:
+  InlineMD(const InlineMD& other) = default;
+
+ public:
+  nlohmann::json to_json() const final {
+    nlohmann::json j;
+    j["markedInline"] = markedInline;
+    j["likelyInline"] = likelyInline;
+    j["isTemplate"] = isTemplateFunction;
```
TimHeldmann, line 35: While I think it is worth to remember whether a function is a template,
I wonder if `InlineMD` is a good place for this information to be.

I do acknowledge, that we are probably constrained by the exported `InlineMD` of cgc.
Any reasonable way to change this?

graph/include/metadata/InlineMD.h
```diff
@@ -0,0 +1,75 @@
...
+    j["markedAlwaysInline"] = markedAlwaysInline;
```
TimHeldmann, line 36: Aren't these mutually exclusive?
Could we just have an enum (or similar) like:
`inlineStatus:[markedInline|likelyInline|markedAlwaysInline]`

TimHeldmann (commented):
No technical problems, only questions for clarification and suggestions for further thought

graph/include/metadata/InlineMD.h
sebastiankreutzer: I don't think we can assume they are. For example, you can mark templates `inline` or have both `inline` and the `always-inline` attirbute.

graph/include/metadata/InlineMD.h
sebastiankreutzer: Fully agree with you. I think there is a lot of old baggage when it comes to how the current metadata is structured. 
For now I would suggest leaving it as is and instead, plan a big metadata overhaul in the future.

graph/include/metadata/FileMD.h
```diff
@@ -0,0 +1,63 @@
...
+      this->fromSystemInclude = toMergeDerived->fromSystemInclude;
+    }
+    // TODO: Save all paths, if multiple definitions? For now, just keep the current data.
+  }
+
+  MetaData* clone() const final { return new FilePropertiesMD(*this); }
+
+  std::string origin;
+  bool fromSystemInclude;
+  int lineNumber;
```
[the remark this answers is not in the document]
sebastiankreutzer: Yeah, I left it in there because it was in the old implementation.

graph/include/metadata/FileMD.h
```diff
@@ -0,0 +1,63 @@
...
+      metacg::MCGLogger::instance().getErrConsole()->error(
+          "The MetaData which was tried to merge with FilePropertiesMD was of a different MetaData type");
+      abort();
+    }
+    const FilePropertiesMD* toMergeDerived = static_cast<const FilePropertiesMD*>(&toMerge);
+    if (this->origin.empty()) {
+      this->origin = toMergeDerived->origin;
+      this->fromSystemInclude = toMergeDerived->fromSystemInclude;
+    }
+    // TODO: Save all paths, if multiple definitions? For now, just keep the current data.
```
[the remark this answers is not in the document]
sebastiankreutzer: Good to know!

sebastiankreutzer:
I removed `FileMD,cpp` and instead erase the `fileProperties` JSON entry in the conversion function, after it has been read.
This gets rid of all the warnings.

graph/include/metadata/InlineMD.h
TimHeldmann: I see, thank you

sebastiankreutzer:
> Deleting the `fileProperties` does remove this warning. The `fileMD`, however, also carries information about whether it is a system include. Deleting the entry would possibly delete more information than intended.
> 
> I can't remember using this information ever, so I am fine with including this for now.


Good point! I'm actually using the system include property in CaPI, now that I think about it...

Maybe the better idea is then to have a built-in `FilePropertiesMD`, but remove the `origin` field? There may also be more properties we want to add to it at a later point.


> Additionally we should look into switching to only emitting one warning for a missing metadata definition, instead of emitting a warning for every node with this type of metadata attached.

That would be really useful! Especially if there are metadata entries that we do not use in all tools (e.g. `CallCountEstimation`).
Do you have an idea how to do this @TimHeldmann ?

sebastiankreutzer:
Third attempt :smile: 

This re-adds the `FilePropertyMD` but removes the `origin` field. It also erases the `origin` entry from the JSON input in order to avoid mistakenly using it somewhere.

graph/include/metadata/FilePropertiesMD.h
```diff
@@ -0,0 +1,53 @@
+/**
+ * File: FileMD.h
```
jplehr, line 2: requires update

graph/include/metadata/FilePropertiesMD.h
```diff
@@ -0,0 +1,53 @@
...
+ private:
+  FilePropertiesMD(const FilePropertiesMD& other)
+      : fromSystemInclude(other.fromSystemInclude) {}
+
+ public:
+  nlohmann::json to_json() const final {
+    nlohmann::json j;
+    j["systemInclude"] = fromSystemInclude;
+    return j;
+  };
```
jplehr, line 33:
```suggestion
  }
```

graph/include/metadata/InlineMD.h
```diff
@@ -0,0 +1,75 @@
...
+    return j;
+  };
```
jplehr, line 38:
```suggestion
  }
```

graph/include/metadata/InlineMD.h
```diff
@@ -0,0 +1,75 @@
...
+
+  void merge(const MetaData& toMerge) final {
+    if (std::strcmp(toMerge.getKey(), getKey()) != 0) {
+      metacg::MCGLogger::instance().getErrConsole()->error(
+          "The MetaData which was tried to merge with NumStatementMetaData was of a different MetaData type");
+      abort();
+    }
+    const InlineMD* toMergeDerived = static_cast<const InlineMD*>(&toMerge);
+    if (this->isTemplateFunction != toMergeDerived->isTemplateFunction) {
+      metacg::MCGLogger::instance().getErrConsole()->error("Merging functions with mismatched 'isTemplate' metadata");
```
jplehr, line 50: Should this do more than emit an error, or maybe emit a warning instead of an error?
I mostly wonder how harmful this error case is.

sebastiankreutzer: Agreed, I'll change it to a warning

jplehr:
Maybe just as a general comment: Our release builds (which I believe is our default) **do not enable** assertions. So, if you are using them as a safety net, be warned!

graph/include/metadata/InlineMD.h
jplehr: ^

sebastiankreutzer:
> Maybe just as a general comment: Our release builds (which I believe is our default) **do not enable** assertions. So, if you are using them as a safety net, be warned!

Makes sense! In this case, it's not expected that the two metadata types would ever mismatch.

PR #11, opened 2025-04-02
[GH] Use pull_request default events

jplehr:


jplehr:
I think we want to have the default behavior (from [docs](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request)):

> For example, if no activity types are specified, the workflow runs when a pull request is opened or reopened or when the head branch of the pull request is updated.

Previously it would not trigger on all events that we care about.

sebastiankreutzer (approved):
LGTM

PR #13, opened 2025-04-03
Add support for LLVM 19 and 20 (not CI tested)

sebastiankreutzer:
This patch adds support the newest version of LLVM. This is not tested in the CI yet and versions 19 and 20 are therefore not marked as supported.
Instead of giving an error, CMake now only prints a warning, if an  unsupported version is detected.

### DIFF
--- a/cgcollector/lib/include/MetaCollector.h
+++ b/cgcollector/lib/include/MetaCollector.h
@@ -81,16 +81,20 @@ class FilePropertyCollector : public MetaCollector {
     const auto sourceLocation = decl->getLocation();
     auto& astCtx = decl->getASTContext();
     const auto fullSrcLoc = astCtx.getFullLoc(sourceLocation);
+    std::string fileNameStr = "";
+        if (fullSrcLoc.isValid()) {
 #if LLVM_VERSION_MAJOR >= 18
-    const auto fileEntry = fullSrcLoc.getFileEntryRef();
+      const auto fileEntry = fullSrcLoc.getFileEntryRef();
 #else
-    const auto fileEntry = fullSrcLoc.getFileEntry();
+      const auto fileEntry = fullSrcLoc.getFileEntry();
 #endif
-    if (!fileEntry) {
-      return result;
+      if (!fileEntry) {
+        return result;
+      }
+
+      const auto fileName = fileEntry->getName();
+      fileNameStr = fileName.str();
     }
-    const auto fileName = fileEntry->getName();
-    std::string fileNameStr = fileName.str();
 
     result->isFromSystemInclude = astCtx.getSourceManager().isInSystemHeader(sourceLocation);
 
--- a/cgcollector/lib/include/MetaCollector.h
+++ b/cgcollector/lib/include/MetaCollector.h
@@ -82,7 +82,7 @@ class FilePropertyCollector : public MetaCollector {
     auto& astCtx = decl->getASTContext();
     const auto fullSrcLoc = astCtx.getFullLoc(sourceLocation);
     std::string fileNameStr = "";
-        if (fullSrcLoc.isValid()) {
+    if (fullSrcLoc.isValid()) {
 #if LLVM_VERSION_MAJOR >= 18
       const auto fileEntry = fullSrcLoc.getFileEntryRef();
 #else
--- a/cgcollector/lib/include/MetaCollector.h
+++ b/cgcollector/lib/include/MetaCollector.h
@@ -81,7 +81,11 @@ class FilePropertyCollector : public MetaCollector {
     const auto sourceLocation = decl->getLocation();
     auto& astCtx = decl->getASTContext();
     const auto fullSrcLoc = astCtx.getFullLoc(sourceLocation);
+#if LLVM_VERSION_MAJOR >= 18
+    const auto fileEntry = fullSrcLoc.getFileEntryRef();
+#else
     const auto fileEntry = fullSrcLoc.getFileEntry();
+#endif
     if (!fileEntry) {
       return result;
     }
--- a/graph/include/CgNode.h
+++ b/graph/include/CgNode.h
@@ -100,13 +100,13 @@ class CgNode {
    */
   template <typename T, typename... Args>
   T* getOrCreateMD(const Args&... args) {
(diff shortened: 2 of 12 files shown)


sebastiankreutzer:
Interesting, my local GCC and Clang builds where happy. Good thing we have the CI!

sebastiankreutzer:
The last fixups re-add the `template` specifier where it's definitely needed (at least for older GCC versions), i.e. in front of function template invocations with dependent parameters. Took a few tries, but this should work for all versions now.

pgis/lib/include/MetaData/CgNodeMetaData.h
```diff
@@ -240,12 +240,12 @@ class PiraOneData : public metacg::MetaData::Registrar<PiraOneData> {
...
-   data->setNumberOfStatements(numStmts);
-   data->setHasBody(hasBody);
-   data->setDominantRuntime(dominantRuntime);
-   data->setComesFromCube(inPrevProfile);
+  const auto& [has, data] = node->template checkAndGet<PiraOneData>();
+  if (has) {
+    data->setNumberOfStatements(numStmts);
+    data->setHasBody(hasBody);
+    data->setDominantRuntime(dominantRuntime);
+    data->setComesFromCube(inPrevProfile);
```
TimHeldmann, line 248: Is this a formatting error or correction?

graph/include/metadata/MetaData.h
```diff
@@ -36,8 +36,8 @@ class MetaDataFactory {
   template <class... T>
   static CRTPBase* create(const std::string& s, const nlohmann::json& j) {
     if (data().find(s) == data().end()) {
-      MCGLogger::instance().getErrConsole()->template warn(
-          "Could not create: {}, the Metadata is unknown in you application", s);
+      MCGLogger::instance().getErrConsole()->warn("Could not create: {}, the Metadata is unknown in you application",
+                                                  s);
```
TimHeldmann, line 40: I would prefer to have the error in one single line, instead of having only the formatting parameter in a new line.
```suggestion
      MCGLogger::instance().getErrConsole()->warn(
          "Could not create: {}, the Metadata is unknown in you application",  s);
```

TimHeldmann (commented):
Looks good to me.
It is a bit unlucky we have to resort to more and more version specific code.
But I think it is still to little to be a problem.

jplehr:
> Looks good to me.
> 
> It is a bit unlucky we have to resort to more and more version specific code.
> 
> But I think it is still to little to be a problem.

Agreed! Also, that's what you need to do when you do not have a stable API. And I still prefer this over using the (more stable) C interface.

If this increases more substantially, we should reflect on whether we care enough and if we want to add a wrapper layer.

jplehr (changes requested):
Please run a git-clang-format again

graph/include/CgNode.h
```diff
@@ -106,7 +106,7 @@ class CgNode {
     }
 
     auto nmd = new T(args...);
-    this->template addMetaData(nmd);
```
jplehr, line 109: I would think this is a change that is not related to the ability to use newer llvm versions. Am I missing something?
Will this break some GCC compatibility with older versions or so?

sebastiankreutzer: Yes, it is related. Using the `template` keyword results in the following:
```
/home/sebastian/git/MetaCG-Gitlab/graph/include/CgNode.h:109:20: error: a template argument list is expected after a name prefixed by the template keyword [-Wmissing-template-arg-list-after-template-kw]
  109 |     this->template addMetaData(nmd);
```
This works in previous LLVM and GCC versions, but is not standard compliant. Since LLVM 19 (?) this is handled more strictly and results in the error shown above.

pgis/lib/include/MetaData/CgNodeMetaData.h
sebastiankreutzer: This is a correction. I touched the file at some point so it was auto-formatted.
I think I reverted that particular change, so the content is the same. 
I would just leave this bit of formatting in because I think clang-format will complain otherwise.
Does this sound ok?

graph/include/metadata/MetaData.h
sebastiankreutzer: I'm inclined to agree. Unfortunately clang-format does not and automatically changes it back to the original version :smile:

sebastiankreutzer:
I think I addressed all the feedback and re-formatted the sources. Could you have another look, @jplehr and @TimHeldmann ?
Thanks!

graph/include/CgNode.h
jplehr: So this is enabling the compilation with Clang 19 and beyond?
Don't want to block it, just trying to understand as I always thought that this is about getting the API calls right, not something else.

cgcollector/lib/include/MetaCollector.h
```diff
@@ -81,12 +81,20 @@ class FilePropertyCollector : public MetaCollector {
     const auto sourceLocation = decl->getLocation();
     auto& astCtx = decl->getASTContext();
     const auto fullSrcLoc = astCtx.getFullLoc(sourceLocation);
-    const auto fileEntry = fullSrcLoc.getFileEntry();
-    if (!fileEntry) {
-      return result;
+    std::string fileNameStr = "";
```
jplehr, line 84: I think this can remain at the original position.

pgis/lib/include/MetaData/CgNodeMetaData.h
jplehr: Yes, leave it in.

jplehr (commented):
Some nits. I'll leave it with @TimHeldmann to accept.

graph/include/CgNode.h
sebastiankreutzer: It's both - enabling the CGCollector to work with Clang 19+ and compiling the project as a whole with thise Clang versions.

jplehr: How is this particular change with the `template` keyword related to enable CGCollector to work with Clang 19+? That's the bit that I do not understand.
I can see how this enables to compile CGCollector with Clang 19+, but not how this impacts to use Clang19+ as a library.

PR #14, opened 2025-04-04
Add unique logging output capability with tests

TimHeldmann:
This PR shows a demo implementation for outputting unique error messages.
It adds wrappers around all spdlog logging functions.
Usability is still suboptimal, as you need to be very verbose.

I am willing to add additional utility functions like:
`info_c_u(...)` or `info_console_unique(...)` so one does not have to write:
`info<metacg::MCGLogger::Output::DEFAULT, metacg::MCGLogger::LogType::UNIQUE>(...)`

I am looking forward to your feedback


graph/include/LoggerUtil.h
```diff
@@ -32,11 +34,108 @@ class MCGLogger {
    */
   inline auto getErrConsole() { return errconsole.get(); }
 
+  enum class LogType { DEFAULT, UNIQUE };
+  enum class Output { DEFAULT, ERROR };
+
+  template <Output outPutType=Output::DEFAULT, LogType lt = LogType::DEFAULT, typename MSG_t, typename... Args>
```
jplehr, line 40: Would it make sense the switch the `Output` and the `LogType` template parameters?
That way, we would have only one parameter to be explicit about if we wanted to uniquify the output, but use the general output sink.
IMHO, we would then define `error` and `warn` to go to `stderr` and the other go to `stdout` by default.
What do others think?

TimHeldmann: I don't have a strong opinion on this, but currently there is no consistent use of the different streams.
Stdout has all kinds of errors, warning and infos.
Atleast stderr is only used for errors and warnings as far as I can tell.

If we decide to have errors and warnings go to stderr by default we should make another pass over the codebase to get this right.

jplehr: Agreed! We also send errors to the user, but continue execution as if nothing happened. So.. maybe we should turn all these into warnings at the same time.

jplehr:
This btw also needs a git-clang-format, I think.

graph/test/unit/MCGManagerTest.cpp
```diff
@@ -6,8 +6,6 @@
 
 #include "gtest/gtest.h"
 
-#include "LoggerUtil.h"
```
jplehr, line 9: Unrelated NFC change I think

graph/test/unit/ReachabilityAnalysisTest.cpp
```diff
@@ -6,7 +6,6 @@
 
 #include "gtest/gtest.h"
 
-#include "LoggerUtil.h"
```
jplehr, line 9: Unrelated NFC change

jplehr (commented):
There are a few more changes that should be another NFC as it's unrelated to the functional change in this PR

graph/include/LoggerUtil.h
```diff
@@ -32,11 +34,108 @@ class MCGLogger {
...
+      getErrConsole()->trace(formattedMessage);
+    }
+  };
+
  private:
   MCGLogger() : console(spdlog::stdout_color_mt("console")), errconsole(spdlog::stderr_color_mt("errconsole")) {}
 
+
+  template <LogType lt>
+  inline bool uniqueify(const std::string& formattedMessage) {
```
jplehr, line 123:
```suggestion
  inline bool ensureUnique(const std::string& formattedMessage) {
```
IMHO the word `uniqueify` reads a bit weird.

jplehr:
I think this should go ahead and implement the suggested change. As a follow-up the new logging capability can be used to address the mentioned inconsistencies in where messages are going. Also as a follow-up the suggested free-standing convenience functions can be implemented.

graph/include/LoggerUtil.h
TimHeldmann: Changed function name to suggested

jplehr (commented):
Thank you for moving this forward. I have couple of more comments.

graph/test/unit/MCGManagerTest.cpp
jplehr: Since there are so many NFC changes to the tests in this PR anyway, keep it in.

graph/include/LoggerUtil.h
```diff
@@ -32,11 +34,232 @@ class MCGLogger {
...
+  /**
+   * A shorthand to print a unique info to the console
+   *
+   * @tparam MSG_t inferred from msg
+   * @tparam Args  inferred from args
+   * @param msg the msg to be formatted and printed
+   * @param args the arguments to format the message with
+   */
+  template <typename MSG_t, typename... Args>
+  void info_c_u(const MSG_t msg, Args&&... args) {
```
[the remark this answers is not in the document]
sebastiankreutzer: I agree with this. However, I suggest putting the message kind first, i.e. `logInfoUnique` and `logErrorUnique` for the unique versions, and `logInfo` and `logError` for "normal" logging.

graph/include/LoggerUtil.h
```diff
@@ -32,11 +34,232 @@ class MCGLogger {
    */
   inline auto getErrConsole() { return errconsole.get(); }
 
+  enum class LogType { DEFAULT, UNIQUE };
+  enum class Output { DEFAULT, ERROR };
```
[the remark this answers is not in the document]
sebastiankreutzer: I agree with this, at least `DEFAULT` should be renamed to clarify it corresponds to `stdout`. This also avoid confusion with the `LogType`.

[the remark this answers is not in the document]
jplehr: Maybe not `stdout` and `stderr` but `StdConsole` and `ErrConsole`? Maybe we do something funky so that it is not "just stdout"?

graph/include/LoggerUtil.h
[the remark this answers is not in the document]
TimHeldmann: How do we differentiate between the two output consoles?
`logErrorUnique` vs `errErrorUnique` or do you prefer `logErrorUnique` vs `logErrorUniqueErrConsole` ?

graph/include/LoggerUtil.h
```diff
@@ -32,11 +34,232 @@ class MCGLogger {
...
+   * A shorthand to print a unique info to the console
+   *
+   * @tparam MSG_t inferred from msg
+   * @tparam Args  inferred from args
+   * @param msg the msg to be formatted and printed
+   * @param args the arguments to format the message with
+   */
+  template <typename MSG_t, typename... Args>
+  void info_c_u(const MSG_t msg, Args&&... args) {
+    info<LogType::UNIQUE, Output::DEFAULT>(msg, std::forward<Args>(args)...);
```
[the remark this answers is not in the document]
TimHeldmann: We could have this be picked up by the defaults in the function definition, but I decided to be explicit here, for clarity.

graph/include/LoggerUtil.h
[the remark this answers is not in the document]
TimHeldmann: I agree `StdConsole` and `ErrConsole` are significantly better names.

graph/test/unit/VersionTwoMCGWriterTest.cpp
```diff
@@ -81,7 +80,7 @@ TEST_F(V2MCGWriterTest, OneNodeCGWrite) {
   auto& mcgm = metacg::graph::MCGManager::get();
   const auto& cg = mcgm.getCallgraph();
   cg->insert("main");
-#pragma warning(suppress: -Wdeprecated-declarations)
+#pragma warning(suppress : -Wdeprecated - declarations)
```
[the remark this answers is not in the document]
TimHeldmann: Yes, changed it back.

graph/include/LoggerUtil.h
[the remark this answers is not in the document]
jplehr: IMHO the convenience functions should only do "the default". For everything more funky, the user needs to be explicit.
So, `err` should always log to `ErrConsole`.

graph/include/LoggerUtil.h
```diff
@@ -32,11 +34,232 @@ class MCGLogger {
...
+  /**
+   * A shorthand to print a unique info to the console
+   *
+   * @tparam MSG_t inferred from msg
+   * @tparam Args  inferred from args
+   * @param msg the msg to be formatted and printed
+   * @param args the arguments to format the message with
+   */
+  template <typename MSG_t, typename... Args>
+  void logInfoUnique(const MSG_t msg, Args&&... args) {
```
jplehr, line 187: Now that I look at it, wasn't the point here to not need to get the logger instance and all of that? @sebastiankreutzer can you comment on that again?

I don't remember if we disagreed to have them (because I said I'd prefer to be explicit about singletons or so), but I thought that the convenience functions would be free-standing and get the logger.
So basically:

```
logInfoUnique(message...) {
  auto logger = metaCG::Logger::instance();
  logger->info<Whatever>( message );
}
```

So, at the call site, all you do is

```
 logInfoUnique("This is my message");
```

TimHeldmann: While I do remember the discussion, I did not remember to implement this.
Good catch, thank you.

graph/include/LoggerUtil.h
```diff
@@ -32,11 +34,245 @@ class MCGLogger {
...
+      }
+      allreadyPrintedMessages.insert(std::hash<std::string>()(formattedMessage));
+    }
+    return false;
+  }
+
   std::shared_ptr<spdlog::logger> console;
   std::shared_ptr<spdlog::logger> errconsole;
+
+  std::unordered_set<size_t> allreadyPrintedMessages;
```
jplehr, line 275:
```suggestion
  std::unordered_set<size_t> alreadyPrintedMessages;
```
Just found that typo during my last pass.

jplehr (approved):
LGTM with one typo nit.

graph/test/unit/LoggingTest.cpp
```diff
@@ -0,0 +1,391 @@
...
+            "[errconsole] [debug] "
+            "Test String 1, 2.3 "
+            "\n");
+}
+
+TEST_F(LoggingTest, WarnConsoleDefault) {
+  auto logger = metacg::MCGLogger::instance();
+  testing::internal::CaptureStdout();
+  logger.warn("Test String {}, {} ", 1, 2.3f);
+  // logger.getConsole()->warn("Test
```
jplehr, line 140: Please remove this leftover

graph/test/unit/LoggingTest.cpp
```diff
@@ -0,0 +1,391 @@
...
+            "[warning] Test "
+            "String 1, 2.3 \n");
+}
+
+TEST_F(LoggingTest, ShorthandWarn) {
+  testing::internal::CaptureStdout();
+  metacg::MCGLogger::logWarn("Test String {}, {} ", 1, 2.3f);
+  std::string output = testing::internal::GetCapturedStdout();
+  removeDate(output);
+  ASSERT_EQ(output, "[console] [info] Test String 1, 2.3 \n");
```
jplehr, line 199: I believe this needs to be 'warn' instead of 'info'

graph/test/unit/LoggingTest.cpp
```diff
@@ -0,0 +1,391 @@
...
+}
+
+TEST_F(LoggingTest, ShorthandWarnUnique) {
+  testing::internal::CaptureStdout();
+  metacg::MCGLogger::logWarnUnique("Test String {}, {} ", 1, 2.3f);
+  metacg::MCGLogger::logWarnUnique("Test String {}, {} ", 1, 2.3f);
+  std::string output = testing::internal::GetCapturedStdout();
+  removeDate(output);
+  ASSERT_EQ(output, "[console] [info] Test String 1, 2.3 \n");
```
jplehr, line 208: Same

graph/test/unit/LoggingTest.cpp
```diff
@@ -0,0 +1,391 @@
...
+}
+
+TEST_F(LoggingTest, ErrorConsoleDefault) {
+  auto logger = metacg::MCGLogger::instance();
+  testing::internal::CaptureStdout();
+  logger.error("Test String {}, {} ", 1, 2.3f);
+  std::string output = testing::internal::GetCapturedStdout();
+  removeDate(output);
+  ASSERT_EQ(output,
+            "[console] [error] "
```
jplehr, line 218: Does this need to be ErrConsole here?

graph/test/unit/LoggingTest.cpp
```diff
@@ -0,0 +1,391 @@
...
+            "Test String 1, 2.3 "
+            "\n");
+}
+
+TEST_F(LoggingTest, ShorthandErr) {
+  testing::internal::CaptureStdout();
+  metacg::MCGLogger::logError("Test String {}, {} ", 1, 2.3f);
+  std::string output = testing::internal::GetCapturedStdout();
+  removeDate(output);
+  ASSERT_EQ(output, "[console] [info] Test String 1, 2.3 \n");
```
jplehr, line 271: ErrConsole here?

graph/test/unit/LoggingTest.cpp
```diff
@@ -0,0 +1,391 @@
...
+}
+
+TEST_F(LoggingTest, ShorthandErrUnique) {
+  testing::internal::CaptureStdout();
+  metacg::MCGLogger::logErrorUnique("Test String {}, {} ", 1, 2.3f);
+  metacg::MCGLogger::logErrorUnique("Test String {}, {} ", 1, 2.3f);
+  std::string output = testing::internal::GetCapturedStdout();
+  removeDate(output);
+  ASSERT_EQ(output, "[console] [info] Test String 1, 2.3 \n");
```
jplehr, line 280: Same

jplehr (commented):
Looked at the check failures and commented on some places that I think are the issues

PR #16, opened 2025-04-14
Remove unnecessary error reporting in cgvalidate tests; fix typo from cgverify to cgvalidate

TimHeldmann:
Cgvalidate uses std::cerr as output stream for errors.
These are not captured in the logfile.
This PR also redirects std::cerr to the logfile in the cgvalidate integration tests.

It also removes the use of the old name cgverify from the startup message in favor of cgvalidate


jplehr (approved):
LGTM

PR #17, opened 2025-04-17
Rework build_submodules.sh and handle different packaged llvm versions

gomfol12:
Reopen PR because development has shifted to this repo.

Original PR-Message:

I had some trouble building this project. So I improved/rewrote the build_submodules.sh script to make it more useful outside the CI-Pipeline. And some distros, like arch, package LLVM differently. So I also updated the cmake files to check for that case.

What I did:

- the script now also downloads and build cubelib
- python venv. Because most modern distros don't allow installing python packages globally through pip anymore.
- updated the container files to new script behavior.
- updated README.md
- tested full-build container file. Works.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+.venv/
+build/
+extern/
--- a/.github/workflows/mcg-ci.yml
+++ b/.github/workflows/mcg-ci.yml
@@ -52,32 +52,32 @@ jobs:
         uses: addnab/docker-run-action@v3
         with: 
           image: metacg-devel:latest
-          run: /opt/metacg/metacg/build/graph/test/unit/libtests
+          run: /opt/metacg/build/graph/test/unit/libtests
       - name: Run PGIS library tests 
         uses: addnab/docker-run-action@v3
         with: 
           image: metacg-devel:latest
-          run: /opt/metacg/metacg/build/pgis/test/unit/pgistests
+          run: /opt/metacg/build/pgis/test/unit/pgistests
       - name: Run target collector tests 
         uses: addnab/docker-run-action@v3
         with: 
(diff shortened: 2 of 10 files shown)


jplehr:
Wow, this is many changes in a single PR. Can we please separate this out into reasonable chunks and separate PRs?

jplehr:
One more question that occurred to me and I don't remember the answer, maybe someone else does: Did the Extra-P build previously pick-up the cubelib from whatever is in the environment, e.g., via a loaded module? Is that still possible or does this change mean that we will always use our own cubelib?

gomfol12:
> Is that still possible or does this change mean that we will always use our own cubelib?

Yes, this has changed. The build_submodules.sh script now builds extra-p with its own cubelib. But, this could be easily made configurable via an option or environment variable.

jplehr:
Please change the title to better reflect what is changed with this patch. I think this is less about the build process but the building of dependencies / prerequisites.

pearzt:
> > Is that still possible or does this change mean that we will always use our own cubelib?
> 
> Yes, this has changed. The build_submodules.sh script now builds extra-p with its own cubelib. But, this could be easily made configurable via an option or environment variable.

As I see it, I think it is fine that we just offer the `build_submodules.sh` script as a way to build all dependencies in order to facilitate setting up a development environment, before we hopefully have a working Spack setup in the not-too-distant future.

pearzt (approved):
I would vote to move forward and merge this. The changes look good to me and I have also tested them locally. I agree with @jplehr that a more descriptive title should be found.

jplehr:
@gomfol12 i assume you need someone else to merge, right?
@pearzt or @TimHeldmann can you help here, please? Thanks!

PR #18, opened 2025-04-18
[MetaCG] Update contribution guidelines

jplehr:
Our previous contribution guidelines were no longer correct as we have moved our development to open GitHub.
I have incorporated how I perceived we worked for the past years and hope that is reflects reality to some degree.

This is a proposal, please engage and point out what is incorrect or should be mentioned differently. This should not be me making policy here.


pearzt (approved):
LGTM. Thanks for taking care of updating this!

PR #19, opened 2025-04-25
Infer destructor calls in CGC

sebastiankreutzer:
From the original [Gitlab PR](https://git.rwth-aachen.de/tuda-sc/projects/metacg/-/merge_requests/190):

```
This patch addresses issues [#119](https://github.com/tuda-sc/projects/metacg/-/issues/119), [#120](https://github.com/tuda-sc/projects/metacg/-/issues/120) and [#121](https://github.com/tuda-sc/projects/metacg/-/issues/121) .
If --capture-ctors-dtors is passed, edges are inserted for:

implicit destructor calls from local variables
base class destructor calls
implicit member destructors
```

As discussed in the PR, the last commit adds the `--no-infer-ctor-dtor-calls` flag to explicitly disable this behavior.


cgcollector/test/input/allCtorDtor/0008.cpp
```diff
@@ -0,0 +1,15 @@
+// Tests if implicit destruction of member variables is properly captured.
+
+struct A {
+  ~A() {}
+};
+
+struct B {
+  ~B(){};  // ~A() should be called here.
```
jplehr, line 8: Please move the comment above the source line.

jplehr (commented):
One question: do we still have test that check the original behavior (meaning the one that does not use the infer logic)?

Others are minor comments.

cgcollector/test/input/allCtorDtor/0003.gtmcg
```diff
@@ -13,11 +13,21 @@
                 "codeStatistics": {
                     "numVars": 0
                 },
+                "estimateCallCount": {
+                    "calls": {},
+                    "codeRegions": {}
+                },
                 "fileProperties": {
-                    "origin": "/opt/metacg/mcg-local/cgcollector/test/input/allCtorDtor/0666.cpp",
+                    "origin": "/home/groups/da_sc/exafoam/metacg_ng/cgcollector/test/input/allCtorDtor/0003.cpp",
```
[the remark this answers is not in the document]
sebastiankreutzer: I renamed the test (added in a previous patch by me) to be more consistent

sebastiankreutzer:
> One question: do we still have test that check the original behavior (meaning the one that does not use the infer logic)?
> 
> Others are minor comments.

Most of the tests run without `capture-ctors-dtors` and are not affected. I can duplicate the tests in the `allCtorDtor` folder to check both behaviors.

cgcollector/lib/src/CallGraph.cpp
```diff
@@ -799,7 +800,8 @@ class CGBuilder : public StmtVisitor<CGBuilder> {
       return;
     }
 
-    if (auto ctor = CE->getConstructor()) {
+    // Ignoring elidable constructor calls here
+    if (auto ctor = CE->getConstructor(); ctor && !CE->isElidable()) {
```
TimHeldmann, line 804: I don't remember whether we talked about this.
Is there a case, for which a manually specified constructor is elidable, and therefore present in the source code but not in the CallGraph?
If so, I don't think we should ignore all elidable constructors.
Thoughts?

cgcollector/test/input/allCtorDtor/0001.gtipcg
```diff
@@ -31,7 +31,9 @@
         "numStatements": 0,
         "overriddenBy": [],
         "overriddenFunctions": [],
-        "parents": []
+        "parents": [
+            "main"
```
TimHeldmann, line 35: I think having separate tests for this is better, than modifying existing tests, especially if we keep the old behavior around for reference.

So yes, please duplicate the test cases which are affected by your changes.

cgcollector/test/input/allCtorDtor/0003.gtmcg
```diff
@@ -13,11 +13,21 @@
...
                     "systemInclude": false
                 },
                 "globalLoopDepth": 0,
+                "inlineInfo": {
```
TimHeldmann, line 25: Is it necessary to have `inlineInfo` and `esimateCallCount` as part of the tests?
Why do they only show now?
I looks to me like this implicitly adds more "things" to be tested for?

jplehr:
> > One question: do we still have test that check the original behavior (meaning the one that does not use the infer logic)?
> 
> > 
> 
> > Others are minor comments.
> 
> 
> 
> Most of the tests run without `capture-ctors-dtors` and are not affected. I can duplicate the tests in the `allCtorDtor` folder to check both behaviors.

I don't think we need all of them, but have some coverage would be desirable IMHO

TimHeldmann (commented):
Thank you for your work.
I agree that this gives a more accurate CG, but still had some questions, but nothing major.

cgcollector/test/input/allCtorDtor/0008.cpp
sebastiankreutzer: done

sebastiankreutzer:
I'm now running the `allCtorDtor` tests twice, once with the old behavior and once with the new. 
To do this, I extended the testing script so that you can provide multiple ground truth variants for different runs of the same test source file.

cgcollector/test/input/allCtorDtor/0003.gtmcg
sebastiankreutzer: No, it's not necessary. These are emitted by default though. I'm actually not sure why there were not there before and why this did not cause issues before :thinking:

cgcollector/lib/src/CallGraph.cpp
sebastiankreutzer: I'm not entirely sure, tbh. I guess it's a possibility, if this is the only point where this constructor would be invoked?

Do we usually add constructors if they are never called?

sebastiankreutzer: Okay, so looking at some tests, we definitely do emit unused constructors and destructors (see `~B()` in `0007.noinfer.gtipcg` for example).
So I don't think this will cause any missing nodes.

TimHeldmann: Ok, then it's fine

PR #20, opened 2025-04-25
Fix CMake command for full build that is provided in our README

pearzt:
This fixes two small mistakes that had sneaked into the CMake command that we provide in our README for a full build of MetaCG.

- Definition of `extinstalldir` must be in a seperate line in order to be picked up by CMake.
- Actually enable compilation of PGIS, not just claim to do so in the comment.


TimHeldmann (approved):
Looks good to me.

PR #21, opened 2025-04-25
Improve CMake setup for linking cubelib

pearzt:
While chasing down a CI failure in our downstream CI, I stumbled on the way how we handle cubelib in CMake. If I'm not mistaken, the current implementation is fishy, because the `find_library` never actually runs (as we always pre-define `CUBE_LIB`). If it did run, the output would be something like `<...>/libcube4.so`, which would then break the `target_link_directories`.

This PR is an attempt to improve this by actually running `find_library` and `find_path` and introducing a new parameter `CUBE_DIR` for the user to specify the cube installation directory. If other's prefer the old way for some reason, I'm happy to close this.


jplehr (approved):
LGTM

PR #23, opened 2025-04-26
[GH] Ignore NFC paths to trigger CI

jplehr:
The GitHub CI can ignore paths that do not add any functionality, e.g. READMEs, when determining whether it should start a CI job.

Unfortunately, it seems that GitHub CI does not yet support YAML anchors. That means we have to duplicate the list of paths-to-ignore.

### DIFF
--- a/.github/workflows/mcg-ci.yml
+++ b/.github/workflows/mcg-ci.yml
@@ -2,7 +2,39 @@ name: MetaCG CI
 
 on: 
   push:
+    paths-ignore:
+      - 'README.md'
+      - 'LICENSE.txt'
+      - 'CONTRIBUTING.md'
+      - 'CITATION.cff'
+      - 'AUTHORS'
+      - 'formatSrc.sh'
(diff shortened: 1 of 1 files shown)


.github/workflows/mcg-ci.yml
```diff
@@ -2,7 +2,41 @@ name: MetaCG CI
...
+    paths-ignore:
+      - 'README.md'
+      - 'LICENSE.txt'
+      - 'CONTRIBUTING.md'
+      - 'CITATION.cff'
+      - 'AUTHORS'
+      - 'formatSrc.sh'
+      - '.gitlab-ci.yml'
+      - '.gitignore'
+      - '.dockerignore'
```
pearzt, line 14: Maybe I'm splitting hairs here, but a change to `.dockerignore` could influence the CI, right? If so, we probably should not include it here.

sebastiankreutzer (approved):
LGTM

pearzt (approved):

PR #24, opened 2025-04-26
[NFC] Fix two README errors

jplehr:
README was wrong in one place and inconsistent in another place.


pearzt (approved):
Oh, thanks for spotting this!

PR #32, opened 2025-05-02
Implement Python bindings `pymetacg`

pearzt:
With this PR, I propose `pymetacg` as Python bindings for MetaCG. It is designed to allow for parsing of MetaCG files and read-only access to callgraphs, including node meta data. The idea is to allow for easy programmatic access to call graphs, for example for the automatic evaluation of tools that produce or process MetaCG call graphs.

`pymetacg` is an opt-in feature that introduces [nanobind](https://github.com/wjakob/nanobind) as an additional dependency. The following additions are made:
- `pymetacg` nanobind module with associated helper/boilerplate code
- Tests to check for functionality of the Python bindings (using Pytest; compilation has to be switched on independently)
- CMake setup to enable compilation of `pymetacg` and the test suite
- `pymetacg/README.md` to explain the capabilities and provide instructions how to enable the feature
- Mention feature in the main README
- Changes to the CI setup to run pymetacg's tests as part of our CI

Please refer to `pymetacg/README.md` for an example how `pymetacg` can be used.

I'm aware that this a lot of new code to review, but I don't see a reasonable way to split this up without merging dead or untested code. I suggest that we discuss the changes in one of our upcoming meetings.


sebastiankreutzer:
This is great! On a first pass, I don't see any glaring issues. I agree that we should discuss this feature in the next meeting before merging, though.

cmake/nanobind.cmake
```diff
@@ -0,0 +1,48 @@
...
+if(CMAKE_VERSION
+   VERSION_LESS
+   3.18
+)
+  set(DEV_MODULE Development)
+else()
+  set(DEV_MODULE Development.Module)
+endif()
+
+find_package(
```
jplehr, line 12: Does this allow to find and use other python versions or is this only compatible with Python 3.8?

pymetacg/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+include(nanobind)
+include(spdlog)
+include(json)
+
+nanobind_build_library(nanobind-static)
+
+# nanobind_add_module(pymetacg NOSTRIP pymetacg.cpp) We use nanobind's low-level API, as the high-level API sets the
```
jplehr, line 7: Leftover from experimenting? 
Comment is very good, just formatting seems a bit off

pymetacg/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+
+add_metacg_library(pymetacg)
+add_spdlog_libraries(pymetacg)
+add_json(pymetacg)
+
+if(NOT CMAKE_INSTALL_PYTHON_LIBDIR)
+  set(CMAKE_INSTALL_PYTHON_LIBDIR
+      "${CMAKE_INSTALL_LIBDIR}/python${Python_VERSION_MAJOR}.${Python_VERSION_MINOR}/site-packages"
+  )
+  # ${Python_SITELIB})
```
jplehr, line 26: Leftover?

pymetacg/README.md
```diff
@@ -0,0 +1,92 @@
+# pymetacg
+
+`pymetacg` are MetaCG's Python bindings that support parsing MetaCG files from Python projects.
+Currently, manipulating and writing call graphs is *not* supported.
+
+## Dependencies
+ - [nanobind](https://github.com/wjakob/nanobind)
+ - [pytest](https://pypi.org/project/pytest/) (for tests only)
+ - [pytext-cmake](https://pypi.org/project/pytest/) (for tests only)
```
jplehr, line 9: Typo and link wrong

jplehr (commented):
Thanks for putting this up

Only initial drive by comments.

cmake/nanobind.cmake
pearzt: This is the minimum version, so yes, it works with newer versions. Tbh, this was not a very deliberate choice and I haven't tested it with Python versions other than the ones installed in the CI and my machine. In principle, `pymetacg` should work with all non-ancient Python versions, I think.

pymetacg/CMakeLists.txt
pearzt: Fixed

pymetacg/README.md
pearzt: Fixed

CMakeLists.txt
```diff
@@ -129,3 +129,28 @@ option(
   "On or Off"
   OFF
 )
+
+# Build MetaCG's Python interface
+option(
+  METACG_BUILD_PYMETACG
+  "On or off"
+  OFF
+)
```
jplehr, line 138: More as a note to me and for the future: We should probably think again which CMake options should be `CACHE` variables.

pymetacg/README.md
```diff
@@ -0,0 +1,92 @@
...
+$> cmake --install build
+```
+
+By default, nanobind is downloaded automatically during configruation.
+To use an already installed nanobind, use `-DMETACG_USE_EXTERNAL_NANOBIND=ON` and specify the installation path using `-Dnanobind_ROOT=<your nanobind installation>`, if necessary.
+
+The CMake setup of `pymetacg` uses CMake's [FindPython](https://cmake.org/cmake/help/latest/module/FindPython.html) to find the Python installation for which `pymetacg` will be built.
+To specify which Python installation to use, use `-DPython_ROOT_DIR=<python installation>`.
+
+By default, the resulting dynamic library will be installed in `<installation prefix>/lib/<python version>/site-packages`.
```
jplehr, line 27: What is the `<installation prefix>` mentioned here?

pymetacg/README.md
```diff
@@ -0,0 +1,92 @@
...
+To override the installation path, use `-DCMAKE_INSTALL_PYTHON_LIBDIR`.
+Finally, make sure that the installation directory in found by Python, e.g., by using the `PYTHONPATH` environment variable.
```
jplehr, line 29:
```suggestion
Finally, make sure that the installation directory is found by Python, e.g., by using the `PYTHONPATH` environment variable.
```

pymetacg/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+# read MetaCG file (.mcg)
+cg = pymetacg.Callgraph.from_file("/path/to/callgraph.mcg")
+
+# check whether node exists in graph
+main_exists = "main" in cg
+
+# access nodes in graph
+main = cg["main"]
+foo = cg["main"]
```
jplehr, line 51:
```suggestion
foo = cg["foo"]
```

pymetacg/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+# iterate over nodes in graph and print function names
+for node in cg:
+    print(node.function_name)
+
+# check if node has some meta data
+main_has_statement_metadata = "numStatements" in main.meta_data
+
+# print node meta data
+print(main.meta_data["numStatements"].data)
```
jplehr, line 70: Is this always `.data`?

pymetacg/README.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+## Tests
+`pymetacg` comes with a PyTest-based test suite.
+Use `-DMETACG_BUILD_PYMETACG_TESTS=ON` and install `pytest` and `pytest-cmake` need to be installed from PyPI to enable building to the tests:
```
jplehr, line 75:
```suggestion
The test suite requires `pytest` and `pytest-cmake` to be installed from PyPI.
```

pymetacg/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+# check if node has some meta data
+main_has_statement_metadata = "numStatements" in main.meta_data
+
+# print node meta data
+print(main.meta_data["numStatements"].data)
+```
+
+## Tests
+`pymetacg` comes with a PyTest-based test suite.
```
jplehr, line 74:
```suggestion
`pymetacg` comes with a PyTest-based test suite, which can be enabled via the CMake option `-DMETACG_BUILD_PYMETACG_TESTS=ON`.
```
Text is really just a suggestion. Not sure what's best.

pymetacg/nanobind_json.h
jplehr, line 1: Is it OK to include this file into the project?

pymetacg/pymetacg.cpp
```diff
@@ -0,0 +1,83 @@
+
```
jplehr, line 1: Misses any sort of License / Project header.

pymetacg/util.h
```diff
@@ -0,0 +1,79 @@
+#pragma once
```
jplehr, line 1: Misses License / Project header

pymetacg/util.h
```diff
@@ -0,0 +1,79 @@
...
+
+#include <Callgraph.h>
+#include <CgNode.h>
+#include <CgNodePtr.h>
+#include <MCGManager.h>
+
+using namespace metacg;
```
jplehr, line 8: Not sure if this applies here: But `using` in headers is (generally) dangerous.
Is there a reason that this file should not declare all of it inside `namespace metacg` and then have the `.cpp` file use that namespace?

pymetacg/util.h
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Helper type that wraps around CgNode* and keeps additional information about which call graph
+ * the a node is part of.
```
jplehr, line 12:
```suggestion
 * Helper type that wraps around CgNode* and keeps additional information about which call graph
 * the node is part of.
```

pymetacg/util.h
```diff
@@ -0,0 +1,79 @@
...
+  bool operator==(const NodeContainerIteratorWrapper& other) const { return iter == other.iter; }
+  bool operator!=(const NodeContainerIteratorWrapper& other) const { return !(*this == other); }
+
+ private:
+  NodeIterator iter;
+  const Callgraph& graph;
+};
+
+/**
+ * Auxiliary iteartor to iterate over metacg::CgNodeRawPtrUSet and pack CgNode* into CgNodeWrapper
```
jplehr, line 43:
```suggestion
 * Auxiliary iterator to iterate over metacg::CgNodeRawPtrUSet and pack CgNode* into CgNodeWrapper
```

pymetacg/README.md
pearzt: Yes, unfortunately. I've spent a considerable amount of time trying to expose the JSON-data of meta data objects directly (so without going through `.data`), but couldn't quite figure out how to do that with nanobind. I agree that being able to omit `.data` would be nicer, so if anybody knows how to do that, I would be very interested.

pymetacg/nanobind_json.h
pearzt: I would think it is. It's a modified version of a file that's licensed under a permissive license and I've credited the source. I'm not an expert on this, so please let me know if anyone has doubts.

pymetacg/README.md
jplehr: Ok, but it is the same for all `MetaData`? That was my main question.
I think mentioning that somehow would be good for the uneducated like me.

CMakeLists.txt
pearzt: Agreed :+1: 

For now, it's just analogous with the other options.

pymetacg/README.md
pearzt: Is it clearer now?

pymetacg/pymetacg.cpp
pearzt: Fixed

pymetacg/README.md
jplehr: Is this the MetaCG install prefix? If so, maybe we should be explicit.

pymetacg/README.md
jplehr: Yes, thank you

pymetacg/README.md
pearzt: Done

pearzt:
As it just uses the graphlib's reading facilities, it should be agnostic of the MetaCG format version. Tests use MetaCG v2 at the moment as it was just easier to craft example files for that version. Having at least one test that uses MetaCG v3 seems like a good idea, though.

PR #34, opened 2025-05-05
Feat/origin aware callgraph

TimHeldmann:
We introduced the origin attribute to differentiate between identically named functions from different translation units.
However, we did not implement a way to actually use this additional information when querying the callgraph.

This patch adds this functionality.

During development of this feature I concluded, that the option to query all nodes of a given name, regardless or their origin is useful.
There will be a followup patch in which the feature to get all nodes independent of their origin will be added.

Looking forward to your suggestions and comments.


TimHeldmann:
This is currently incompatible with the new CGMerge2 implementation.
I will have a look if it is an easy fix or whether I need to withdraw this PR.

TimHeldmann:
Most of this is adressed in the [format v4](https://github.com/tudasc/MetaCG/pull/49)

PR #35, opened 2025-05-06
Make headers public to help transitively dependent client programs

TimHeldmann:
Client tools currently need to link to spdlog and nlohman themselves as we expose some of their includes in header files.
This changes the linkage status inside cmake to be public to hopefully remove this problem

### DIFF
--- a/cmake/json.cmake
+++ b/cmake/json.cmake
@@ -25,5 +25,5 @@ else()
 endif()
 
 function(add_json target)
-  target_link_libraries(${target} PRIVATE nlohmann_json::nlohmann_json)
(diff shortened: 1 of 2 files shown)


jplehr:
> [...] to be public to hopefully remove this problem

Does it solve the problem or not (at least in a particular test case that you have)?

TimHeldmann:
It does solve the problem in my case.
But I have not checked if it solves every instance of it.

jplehr (approved):
LGTM

jplehr:
You should probably change the title when merging to make clear it is third party lib headers.

PR #36, opened 2025-05-07
Revert "Make headers public to help transitively dependent client programs"

TimHeldmann:
Reverts tudasc/MetaCG#35

Internal testing has shown problems with installing metacg after this patch has landed.
Reverting for now


jplehr (approved):
LGTM

PR #39, opened 2025-05-22
Add tool CGConvert to convert between file format versions

pearzt:
This is pretty much just what the title says. I was looking for such a tool to generate test cases with file format version 3 to pymetacg, didn't find one, so I cobbled this together.


tools/cgconvert/CGConvert.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include "config.h"
+
+#include <iostream>
+#include <fstream>
+
+#include "LoggerUtil.h"
+#include "io/MCGReader.h"
+#include "io/VersionTwoMCGWriter.h"
+
+// This may appear to be unused, but the variables declared here have side effects on the graph lib
```
jplehr, line 16: What does that mean?

tools/cgconvert/CGConvert.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include "metadata/BuiltinMD.h"
+
+using namespace metacg;
+
+int main(int argc, char** argv) {
+  auto console = MCGLogger::instance().getConsole();
+  auto errConsole = MCGLogger::instance().getErrConsole();
```
jplehr, line 23: You could use the new convenience functions from https://github.com/tudasc/MetaCG/commit/bacb429a60320d2e2f31241f1776e733e8e1bac4

tools/cgconvert/CGConvert.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+  console->info("Running metacg::CGConvert (version {}.{})\nGit revision: {}", MetaCG_VERSION_MAJOR,
+                MetaCG_VERSION_MINOR, MetaCG_GIT_SHA);
+
+  if (argc < 3) {
+    errConsole->error("Usage: {} <input_file> <output_file> [output_version]", argv[0]);
+    return EXIT_FAILURE;
+  }
+
+  std::string inputFile = argv[1];
```
jplehr, line 33: For arguments, I'd prefer if we can transition to something like `cxxopts`, which is used in PGIS already.

jplehr (commented):
Initial review.

The cxxopts comment is a wish-thing. I do understand that it may look like overkill for such a simple tool. Please look at it though and make a reasonable decision.

tools/cgconvert/CGConvert.cpp
pearzt: Tbh, I just copied this from `cgmerge2`. Still, it is necessary:  If we don't include the header, the in-tree metadata types are not part of the translation unit and would thus not be included in the `cgconvert` binary. Then, the tool would not "know" these meta data types and would omit them in the output file.

I already had a brief discussion about this with @TimHeldmann a few weeks ago: One can argue that the in-tree meta data types should be defined in a translation unit that gets linked into the graphlib, so you would get access to those by just linking against MetaCG. That's a seperate discussion, though.

tools/cgconvert/CGConvert.cpp
pearzt: Ah, good idea, thanks - done!

jplehr:
> IMO, `cxxopts` is not necessary here due to the simplicity (we also don't use it for `cgmerge(2)`). 

Yeah, I can understand that point of view. I mostly think that having a uniform way of doing things is beneficial. So, IMHO, we should strive towards a more uniform way, i.e., use cxxopts for everything instead of having this arbitrary "this is simple enough" metric. 

Not a blocker, and not the most important thing. Still my point of view.

tools/cgconvert/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+set(PROJECT_NAME CGConvert)
+set(TARGETS_EXPORT_NAME ${PROJECT_NAME}-target)
+
+add_executable(cgconvert CGConvert.cpp)
+
+add_config_include(cgconvert)
+add_json(cgconvert)
+add_metacg(cgconvert)
+add_spdlog_libraries(cgconvert)
+add_clang(cgconvert)
```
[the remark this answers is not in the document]
pearzt: Same here

TimHeldmann (approved):

PR #41, opened 2025-05-23
Fix cgmerge to handle call graphs with null meta field

silas-martens:
`CGMerge` can now handle call graphs where the meta field is null. This is necessary to merge the patch graphs generated by `cgpatch`.


jplehr (changes requested):
Please add a test for this case

PR #46, opened 2025-07-02
Change json and spdlog into public dependencies of the graphlib

sebastiankreutzer:
JSON and spdlog are both included in certain MetaCG headers. Any tool using MetaCG, therefore, requires them to be in the include path.
This patch adds them to the public interface, thus simplifying the setup for downstream tools.

### DIFF
--- a/graph/CMakeLists.txt
+++ b/graph/CMakeLists.txt
@@ -111,8 +111,8 @@ endif()
 
 add_config_include(metacg)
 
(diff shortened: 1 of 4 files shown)


TimHeldmann:
This seems to be related to: https://github.com/tudasc/MetaCG/pull/35
Which was reverted https://github.com/tudasc/MetaCG/pull/36

Are you sure this works this time?

PR #47, opened 2025-07-02
Replaced error with warning when CGPatch is built with llvm < 15

silas-martens:
Disable CGPatch build for LLVM < 15 instead of failing with error


sebastiankreutzer (approved):
LGTM

jplehr (approved):
LGTM

PR #48, opened 2025-07-06
Performance testing

sebastiankreutzer:
This adds performance tests for the graph lib. I wrote this as a weekend project to be able to evaluate the performance differences of the upcoming graphlib re-write.
 As a follow-up, this can be integrated into the CI to catch performance regressions.
 
 Usage: `./perftest <num_nodes> <num_edges> <rng_seed>`

Results for current `devel` on my laptop: 
```
Test results for 10000 nodes and 20000 edges (seed: 463248923)
---------------------------
Test Name         Time (ms)
---------------------------
read V2 format        40736
write V2 format         866
read V3 format          230
write V3 format         580
insert edges            131
insert nodes             31
---------------------------
```

### DIFF
--- a/graph/test/integration/Performance/PerfTester.cpp
+++ b/graph/test/integration/Performance/PerfTester.cpp
@@ -1,0 +1,179 @@
+/**
+ * File: PerfTester.cpp
+ * License: Part of the MetaCG project. Licensed under BSD 3 clause license. See LICENSE.txt file at
+ * https://github.com/tudasc/metacg/LICENSE.txt
+ */
+
+#include "Callgraph.h"
+#include "io/MCGReader.h"
+#include "io/MCGWriter.h"
+
+#include <chrono>
+#include <functional>
+#include <random>
+
+using namespace metacg;
+
+class Timer {
+ public:
+  using Clock = std::chrono::high_resolution_clock;
+  using TP = Clock::time_point;
+  using Duration = Clock::duration;
+
+  explicit Timer() {start();};
+
+  void start() { startTime = Clock::now();}
+
+  void stop() { endTime = Clock::now(); }
+
+  std::chrono::milliseconds getElapsedMillis() {
+    auto dur = endTime - startTime;
+    auto durInMillis = std::chrono::duration_cast<std::chrono::milliseconds>(dur);
+    return durInMillis;
+  }
+
+ private:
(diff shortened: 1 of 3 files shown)


graph/test/integration/Performance/PerfTester.cpp
```diff
@@ -0,0 +1,186 @@
...
+#include <functional>
+#include <random>
+
+using namespace metacg;
+
+class Timer {
+ public:
+  using clock = std::chrono::high_resolution_clock;
+  using tp = clock::time_point;
+  using duration = clock::duration;
```
jplehr, line 21:
```suggestion
  using Clock = std::chrono::high_resolution_clock;
  using TP = clock::time_point;
  using Duration = clock::duration;
```
Don't know if this makes sense, but I would treat them as "types from MCG" and those are capitalized.

graph/test/integration/Performance/PerfTester.cpp
```diff
@@ -0,0 +1,186 @@
...
+  void insertEdges(Callgraph& cg, int num);
+  void write(Callgraph& cg, io::JsonSink& sink, int version);
+  void read(std::unique_ptr<Callgraph>& cg, io::JsonSource& source);
+
+ private:
+  int seed;
+  int numNodes;
+  int numEdges;
+  std::mt19937 rng;
+
```
jplehr, line 68:
```suggestion
```

graph/test/integration/Performance/PerfTester.cpp
```diff
@@ -0,0 +1,186 @@
...
+  using tp = clock::time_point;
+  using duration = clock::duration;
+
+  explicit Timer() {start();};
+
+  void start() { startTime = clock::now();}
+
+  void stop() { endTime = clock::now(); }
+
+  std::chrono::seconds getElapsedSeconds() {
```
jplehr, line 29: Unused, I think can be removed.

jplehr (approved):
Thanks, that's cool stuff. 

For the future: we can also think of generating some long strings as names, since the older formats are showing real slowness with long strings and they are quite common for C++ mangled names.

jplehr:
I forgot one thing: can you update the commit title to be something like 

[Graph] Add initial performance tests

Or something like that? Thank you.

PR #49, opened 2025-07-07
File format v4 and big graph lib re-write

### LINKED ISSUE #42: RFC: Introduce merge policies
This is the result from a discussion that I head earlier together with @sebastiankreutzer.

## Background
Recently, we changed our graph semantics so that not only the function name is used to uniquely identify a node, but also the `origin` field. These two fields are combined using a hash function and the resulting value is then used as the node ID. This was done so that the graphlib is able to distinguish between two static C/C++ functions of the same name that are defined in different translation units. Currently, the call graph merge is implemented so that it compares this ID to determine whether two nodes should be merged together.

## Problem
We realized that the current implementation might not be sufficiently flexible for some cases. Specifically, we might want to merge call graph that have been generated differently. For instance, we might want to merge a call graph that was generated by a tool operating on the LLVM IR level into a callgraph generated by CGCollector (i.e., on the source-code level). This would currently not be possible, as the LLVM-IR-level tool cannot reconstruct the same origin information that is available to a source-code-level tool. Overall, we may want to have a way to relax the strict definition of node equality that is currently implemented for some use cases.

## Suggested solution
We propose to introduce "merge policies". The default behavior would not be changed, but this would allow us to define specific semantics that can be enabled by passing a merge policy to the call-graph-merge function. Such a policy would then define a function with a signature akin to this: `CgNodePtr getCorrectMergeTarget(CgNodePtr toBeMerged, const Callgraph& callgraphToMergeInto)`. Assuming we want to merge some call graph B into another call graph A, the merge function would then call `getCorrectMergeTarget` for every node of B. The policy's implementation of this function would then return the correct node from A that the node should be merged into. The default policy (maybe: `StrictMergePolicy`) would reflect the current behavior by return the node from A that matches both the function name and the origin (and create a new node if necessary). Other policies would then implement their own semantics to make the choice and potentially throw an error if no decision can be made.
Further down the road, we could also have a merge policy (maybe: `ELFMergePolicy` or `SimpleMergePolicy`) that is designed to reflect the merge semantics implemented in the linker. Such a policy would then need to rely on new meta data that contains information about the function, e.g., whether a function is marked as `static` or `inline`.

## Possible implementations
We see two ways to implemented this: We could either define an abstract class `MergePolicy`. Different policies would then be implemented in subclasses and the user would specify a chosen policy by passing an object of this type to the `merge` function. This implemententation would probably facilitate the development of a plugin interface to define custom merge policies in the future. Alternatively, this could also be implemented as a template parameter for the `merge` function as outlined in the pseudo code below:
```
template<typename Policy>
void merge(CG other) {
    auto policy = Policy::create(other);
    policy.getCorrectMergeTarget(toBeMerged);
}
```

sebastiankreutzer:
# File format v4 and big graph library re-write

This merge request introduces format v4, alongside a number of changes in the graph lib implementation.
Major parts of this PR are:
- Introduction of format v4 to replace v3, addressing the remaining issues of v4 and providing a more compact JSON representation than all prior format versions.
- Re-design of the graph library for improved performance and API consistency.
- Introduction of merge policies (see #42) to handle different merge semantics.

I'm happy for any and all feedback, including:
- Overall design and potential limitations
- Bikeshedding of function/type names
- How this may be split up into reviewable chunks

**Note:** This patch is complete (including unit and integration tests), with the following notable exceptions:
- `VersionThreeMCG{Reader,Writer}Test` has not been adjusted to account for format changes after review. I will update this once we agree on the final format. I/O is still tested via the v2 tests (they depend on the v4 reader/writer).
- `CGConverter`and `CGMerge2` are functional. All other tools need to be updated. Required changes are fairly easy to apply. 
- There are some left-over debugging messages (mostly commented-out) that I will remove at a later point, please disregard these.

For performance results, skip to the very end.

## Format v4

Format v4 combines ideas from both v2 and v3 to provide a JSON representation that is both compact and readable.
V4 completely replaces v3, as keeping both versions while maintaining a consistent internal representation is not feasible.
The following example shows the same call graph in v2, v4 and v3 (use horitzontal scroll bar to see everyghing):

<table>
<tr>
<td>
Version 2
<td>
Version 4 (with names as identifiers)
</td>
<td>
Version 3 (to be deprecated)
</td>
</tr>
<tr>
<td>
<pre>
{
    "_CG": {
        "foo": {
            "callees": [],
            "callers": [
                "main"
            ],
            "doesOverride": false,
            "hasBody": true,
            "isVirtual": false,
            "meta": {
                "fileProperties": {
                    "origin": "foo.cpp",
                    "systemInclude": false
                },
                "numStatements": 0
            },
            "overriddenBy": [],
            "overrides": []
        },
        "main": {
            "callees": [
                "foo"
            ],
            "callers": [],
            "doesOverride": false,
            "hasBody": true,
            "isVirtual": false,
            "meta": {
                "fileProperties": {
                    "origin": "main.cpp",
                    "systemInclude": false
                },
                "numStatements": 3
            },
            "overriddenBy": [],
            "overrides": []
        }
    },
    "_MetaCG": {
        "generator": {
            "name": "CGCollector",
            "sha": "9350298c64ad5a9aef50494f73caad08c38c1a51",
            "version": "0.7"
        },
        "version": "2.0"
    }
}
</pre>
</td>
<td>
<pre>
{
    "_CG": {
        "foo": {
            "functionName": "foo",
            "origin": "foo.cpp",
            "callees": {},
            "hasBody": true,
            "meta": {
                "fileProperties": {
                    "systemInclude": false
                },
                "numStatements": 0,
                "overrideMD": {
                    "overriddenBy": [],
                    "overrides": []
                }
            }
        },
        "main": {
            "functionName" : "main",
            "origin": "main.cpp",
            "callees": {
                "foo": null
            },
            "hasBody": true,
            "meta": {
                "fileProperties": {
                    "systemInclude": false
                },
                "numStatements": 3
            }
        }
    },
    "_MetaCG": {
        "generator": {
            "name": "CGCollector",
            "sha": "a12c2adc9482ef0062fd27f104ab4dfd679cb91f",
            "version": "0.2"
        },
        "version": "4.0"
    }
}
</pre>
</td>
<td>
<pre>
{
    "_CG": {
        "edges": [
            [
                [
                    13570296075836900426,
                    6503480063870808269
                ],
                null
            ]
        ],
        "nodes": [
            [
                6503480063870808269,
                {
                    "functionName": "foo",
                    "origin": "foo.cpp",
                    "hasBody": true,
                    "meta": {
                        "fileProperties": {
                            "systemInclude": false
                        },
                        "numStatements": 0,
                        "overrideMD": {
                            "overriddenBy": [],
                            "overrides": []
                        }
                    }
                }
            ],
            [
                13570296075836900426,
                {
                    "functionName": "main",
                    "origin": "main.cpp",
                    "hasBody": true,
                    "meta": {
                        "fileProperties": {
                            "systemInclude": false
                        },
                        "numStatements": 3
                    }
                }
            ]
        ]
    },
    "_MetaCG": {
        "generator": {
            "name": "MetaCG",
            "sha": "2ce9326f676cfbc802b3924f66dfd97576058b30",
            "version": "0.7"
        },
        "version": "3.0"
    }
}
</pre>
</td>
</tr>

</table>

Here are the main changes compared to the previous formats:
-  The generator can choose arbitrary string IDs to refer to each node. These IDs have no relevance for the in-memory representation.
- Call edges are stored in the `callees` field within each node, similar to v2. Each edge can carry metdata (by default `null`).
  - Compared to a global edges container like v4, this improves human readability and reduces storage requirements (the callee is implicit).
  - Callers are not explicitly stored like in v2 to reduce redundancy. This simplifies manual editing, as only one field needs to be modified, and consistency is always maintained.
- There is no dedicated debug format. Instead, the user can choose to use (uniquified) function names as identifiers during export.

To conserve space, integer identifiers can be used instead of function names:
<table>
<tr>
<td>
Version 4 (with names as identifiers)
<td>
Version 4 (with integers as identifiers)
</td>
</tr>
<tr>
<td>
<pre>
{
    "_CG": {
        "foo": {
            "functionName": "foo",
            "origin": "foo.cpp",
            "callees": {},
            "hasBody": true,
            "meta": {
                "fileProperties": {
                    "systemInclude": false
                },
                "numStatements": 0,
                "overrideMD": {
                    "overriddenBy": [],
                    "overrides": []
                }
            }
        },
        "main": {
            "functionName" : "main",
            "origin": "main.cpp",
            "callees": {
                "foo": null
            },
            "hasBody": true,
            "meta": {
                "fileProperties": {
                    "systemInclude": false
                },
                "numStatements": 3
            }
        }
    },
    "_MetaCG": {
        "generator": {
            "name": "CGCollector",
            "sha": "a12c2adc9482ef0062fd27f104ab4dfd679cb91f",
            "version": "0.2"
        },
        "version": "4.0"
    }
}
</pre>
</td>
<td>
<pre>
{
    "_CG": {
        "0": {
            "functionName": "foo",
            "origin": "foo.cpp",
            "callees": null,
            "hasBody": true,
            "meta": {
                "fileProperties": {
                    "systemInclude": false
                },
                "numStatements": 0,
                "overrideMD": {
                    "overriddenBy": [],
                    "overrides": []
                }
            }
        },
        "1": {
            "functionName": "main",
            "origin": "main.cpp",
            "callees": {
                "0": null
            },
            "hasBody": true,
            "meta": {
                "fileProperties": {
                    "systemInclude": false
                },
                "numStatements": 3
            }
        }
    },
    "_MetaCG": {
        "generator": {
            "name": "MetaCG",
            "sha": "9350298c64ad5a9aef50494f73caad08c38c1a51",
            "version": "0.7"
        },
        "version": "4.0"
    }
}
</pre>
</td>
</table> 
In this example, the size reduction is only 16%, but for larger call graphs with long functions names this can be very effective.
  
Potential additional features, not yet implemented:
- Add a `debug` metadata field. This field can be used to export the list of callers for easier look up during manual inspection. This metadata would be ignored when reading, and re-generated when writing, if requested.

## Graph library

The graph library was adjusted to handle the new format and to accommodate a wider variety of use-cases (e.g., @pearzt 's Flip project).

### Node identification and lookup
Nodes are identified by an integer ID unique to the containing call graph. 
This ID is generally not equal to the string identifier used in the JSON file, but is instead assigned by the call graph when creating the node.
It is explicitly allowed to create multiple nodes with the same function name and/or origin, enabling a correct representation of functions with internal linkage (e.g., functions declared `static`).

We maintain a hash map to simplify look-up by name. The function `const NodeList& getNodes(const std::string&)` can be used to get a list of functions matching the given name. `CgNode* getFirstNode(const std::string&)` is added as a convenience function to look up a single node, e.g. if we know that only one such node exists.

### Node creation
If there are two or more in-memory call graphs, nodes in these call graphs may have identical IDs. 
To avoid issues with unclear ownership and validity of a node, nodes must therefore always be tied to a specific call graph.
This was implemented by making the `CgNode` constructor private. Nodes can now only be created by calling `Callgraph::insert` or `Callgraph::getOrInsertNode`. 
This also guarantees that there are no nodes with invalid or uninitialized IDs, as the IDs are set in the constructor and are marked `const`. 

### Edges
There are no major changes w.r.t. edge representation. Edges are stored centrally in the call graph, as before. 
Instead of function names/hashes, edges are defined by integer node IDs. There is a convenience function `bool addEdge(const std::string&, const std::string&)` to simplify edge creation by name. 

### I/O
Unlinke in v3, there is no direct mapping between the JSON format and the internal representation. 
This makes the implementation of the reader and writer more involved, but allows us to have full control over the import/export process. This also made it possible to provide more helpful error messages, when there are errors in the input JSON.

To convert between the string identifiers in the format and the internal node IDs, mapping classes have been introduced (`StrToNodeMapping` for the reader, `NodeToStrMapping` for the writer). These are used by the reader/writer to maintain a consistent mapping between the two identifiers and to translate node references (e.g. in the `edge` field).
This mapping data structure is only required during reading/writing and is discarded afterward. 
The only use outside the reader/writer itself is in the creation and export of metadata (see below).

### Merging
The implementation of the call graph merge functionality received a major overhaul. 
In the current upstream version, functions are always merged by name and there is no customization of merge semantics. 
This patch introduces so-called merge policies to enable greater control. 

These are based on the following model:

**Merge model**
When merging graph `A` into graph `B`, the merge transfers all nodes from `A` into `B`.
For each node in `A`, there are two main cases, how this can happen:
1. Node does not match any node in `B`: node can be cloned and inserted as a new node in `B`.
2. Node does match a node in B: There are two possible actions: 
   2.1 Keep existing node in `B`: nothing to be done
   2.2 Replace with node from `A`: keep the existing node in `B` but overwrite with properties from `A`  

Merge policies are introduced to concretize the semantics of this model.
Relevant declarations:
```
struct MergeAction { // abbreviated definition
  NodeId targetNode;
  bool replace;
};

struct MergePolicy {
  virtual std::optional<MergeAction> findMatchingNode(const Callgraph& targetCG, const CgNode& sourceNode) const = 0;
};
```
For each node in the source graph `A`, the policy defines which of the outlined cases applies. The function returns:
- An empty value, if there is no matching node (case `1`)
- An action with `replace=false`, if there is a match and the existing node should be kept (case `2.1`).
- An action with `replace=true`, if there is a match and the existing node should be overwritten (case `2.1`).

The policy is passed to `Callgraph::merge`, which executes the merge accordingly. This returns a `MergeRecorder` object, used to look up the performed actions and to map from node IDs from graph `A` into the merged graph `B`. 

Note that metadata are in charge of defining their own merge semantics for cases `2.1` and `2.2`. More details below.

The current implementation comes with the `MergeByName` policy, which models the behavior of the previously used merge strategy. 
As a follow-up, I suggest we add an `ELFMerge` policy that models the actual linking of object files more closely. To do this, we need a `linkage` metadata with values `external`, `internal` and `weak`, and define the policy based on this linkage type. This will finally allow us to accurately handle merge semantics of `static` and `inline` functions, regardless of using a source or IR level generator. 

### Metadata
There are no fundamental changes to how metadata is registered and instantiated.
However, the changes to the format and the merge semantics required some adjustments.

**I/O**
When creating from JSON, the metadata needs to know how to map from string identifiers to internal node IDs. Therefore, the constructor of each metadata takes an additional `StrToNodeMapping&` parameter. This is only relevant if the metadata references other nodes (like in `OverrideMD`) and can be ignored in the majority of cases.

Similarly, the conversion to JSON requires a `NodeToStrMapping` that is now added as a parameter to `toJson`.

**Mapping**
When cloning a node from graph `A` into graph `B`, any node IDs referenced in the metadata become potentially invalid. 
We added the `applyMapping(const GraphMapping&)` method that lets the metadata correctly update its internal node references.
For the majority of metadata that do not reference other nodes, this method can be left empty.

**Merging**
Previously, what should happen during a merge of two metadata objects was often not clear, as there was no information about the performed action.
Now, the executed `MergeAction` is passed as a parameter to allow the metadata to handle the merge accordingly.
Additionally, we pass a `GraphMapping` to map node references of the metadata object that is merged in. Again, this is only relevant if the metadata itself references other nodes.

Please refer to `OverrideMD.h` as an example for a non-trivial metadata implementation using mappings.

## Performance results

The following is a performance comparison with `devel` (both measured on my laptop).

| Test Name       | `devel` (ms) | This PR (ms) | Speedup (×) |
| --------------- | ------------- | ------------- | ----------- |
| insert edges    | 148           | 109           | **1.36×**   |
| insert nodes    | 25            | 15            | **1.67×**   |
| read V2 format  | 28,547        | 457           | **62.45×**  |
| write V2 format | 754           | 735           | **1.03×**   |
| read V3/4 format  | 211 (V3)      | 233 (V4)      | **0.91×**   |
| write V3/4 format | 492 (V3)      | 417 (V4)      | **1.18×**   |

Closes #42


sebastiankreutzer:
As a further note: I would like to completely get rid of the `MCGManager`, as it doesn't have any real purpose anymore beside a `string` → `Callgraph` lookup.
If there are no objections, I will do this after this PR has gone through.

sebastiankreutzer:
The last commit adds methods to delete nodes and edges

graph/src/Callgraph.cpp
```diff
@@ -73,268 +78,318 @@ void Callgraph::clear() {
...
-  return getEdgeMetaData({func1->getId(), func2->getId()}, metadataName);
-}
-MetaData* Callgraph::getEdgeMetaData(const std::pair<size_t, size_t> id, const std::string& metadataName) const {
-  return edges.at(id).at(metadataName);
-}
-MetaData* Callgraph::getEdgeMetaData(const std::string& func1, const std::string& func2,
-                                     const std::string& metadataName) const {
-  return getEdgeMetaData({nameIdMap.at(func1), nameIdMap.at(func2)}, metadataName);
+MetaData* Callgraph::getEdgeMetaData(std::pair<size_t, size_t> ids, const std::string& metadataName) const {
+  assert(edges.find(ids) != edges.end() && "Edge does not exist");
```
jplehr, line 370: Should this be an actual condition or is it OK to have this just as an assert?

graph/src/io/MCGReader.cpp
```diff
@@ -16,8 +16,8 @@ std::unique_ptr<MetaCGReader> createReader(ReaderSource& src) {
   auto versionStr = src.getFormatVersion();
   if (versionStr == "2" || versionStr == "2.0") {
     return std::make_unique<VersionTwoMetaCGReader>(src);
-  } else if (versionStr == "3" || versionStr == "3.0") {
-    return std::make_unique<VersionThreeMetaCGReader>(src);
+  } else if (versionStr == "4" || versionStr == "4.0") {
```
jplehr, line 19: IMHO we should (here or wherever) print an error here when we find version 3 and tell that it has been removed in favor of version 4.

jplehr (commented):
Some initial drive-by comments.

graph/include/io/VersionFourMCGWriter.h
```diff
@@ -0,0 +1,37 @@
...
+#ifndef METACG_VERSIONFOURMCGWRITER_H
+#define METACG_VERSIONFOURMCGWRITER_H
+
+#include "MCGWriter.h"
+#include "config.h"
+namespace metacg::io {
+
+class VersionFourMCGWriter : public MCGWriter {
+ public:
+  // Fixme: Make the MetaCG versions, not CGCollector versions
```
jplehr, line 16: Is this still valid?

graph/src/Callgraph.cpp
sebastiankreutzer: I can change it to a condition. Makes it harder to use wrongly.

PR #50, opened 2025-07-10
Remove support and downstream testing for GCC 10

pearzt:
GCC 10 is around 5 years old by now and, as far as we're currently aware, not a default compiler for any major distribution. Currently, we still test MetaCG using GCC 10 in our downstream CI, but the build are currently failing due a template resolution problem in pymetacg. As briefly discussed during one of our past meetings, I propose to drop support for GCC 10, but would be open to expanding the range of GCC versions again when we move the CI to the new module system on Lichtenberg.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -17,9 +17,6 @@ stages:
 .job-setup: &job-setup
   parallel:
     matrix:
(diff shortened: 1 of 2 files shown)


jplehr (approved):
LGTM

PR #51, opened 2025-07-10
Remove PGIS

pearzt:
PGIS, which is part of the [PIRA](https://github.com/tudasc/PIRA) tool set, was the cradle of the MetaCG project. Although PIRA is currently no longer used or developed, we've still kept PGIS in the MetaCG repository to act as an example use case for the library. Nevertheless, with this PR, I propose to remove PGIS from the repository, mainly because it slows down our development of the MetaCG library and associated tools. Keeping PGIS in the repository would mean that we would have to maintain the code through major changes (like those proposed in #49), without gaining any substantial value from it.

I agree that it's a good idea to present potential users of MetaCG with an example tool, but I would argue that PGIS so longer serves this function anyway as it only makes sense in the PIRA context. IMHO, I would be better to point users to other tools that use MetaCG, e.g., @sebastiankreutzer's [CaPI](https://github.com/tudasc/CaPI) or one of @TimHeldmann's source-to-source tools.

The change would not effect the existing versions of PIRA, as they point to a fixed release of MetaCG (v0.6.2, I think) which still includes PGIS.


README.md
sebastiankreutzer, line 1: I think it would be good to have a note somewhere in here stating that PGIS is no longer maintained and pointing to the last release that includes it.

sebastiankreutzer (commented):
Thanks for preparing this patch!

README.md
pearzt: Good idea, done

jplehr:
I disagree with this change, but it has come up often enough by now that I believe I should not be the blocker.

I think that we should deprecate the tool first, despite it likely not being used outside.
I am a firm believer that we should not only link to projects that use it, but be pretty sure that we have some test coverage for _intended_ uses and I don't think we are there yet.

I do understand that we do not want to have legacy ballast that slows down development significantly, but I also do not want to introduce changes that make the library worst.

jplehr:
Let me be clear: I don't necessarily mean to update PGIS. But maybe, we should just have better tests that do demonstrate intended usages of certain APIs etc.

PR #52, opened 2025-07-10
Remove container build from downstream CI

pearzt:
The container job in our downstream CI has never really worked reliably, mostly due to the shared file system and other infrastructure instabilities. Now that we have container-based testing with the GitHub-CI, I propose to remove this wonky part of the downstream CI.


jplehr (approved):
LGTM

PR #55, opened 2025-07-11
[CI] Add install test to github CI

sebastiankreutzer:
Adds the existing install test to the github CI to ensure that the library is installed correctly.


jplehr (approved):
LGTM

In case the cmake target does some sort of checking that things are actually installed correctly

sebastiankreutzer:
> LGTM
> 
> In case the cmake target does some sort of checking that things are actually installed correctly

The CMake target builds a small test tool that uses the library to read and write a graph. This is the test that was previously failing in the gitlab CI.

PR #67, opened 2025-08-05
Add support for global metadata

sebastiankreutzer:
This patch adds support for attaching global metadata to a call graph.
This is metadata that doesn't belong to one specific node, but to the graph as a whole. 

The implementation requires some slight changes to the v4 file format. Despite this, @pearzt and I came to the conclusion that it is preferable to not increase the format minor version, since v4 has not had an official release.
Inside the graph lib, metadata management is now implemented in a mixin class that is used by both the `CgNode` and `Callgraph` classes.

As an initial use case, I extended the `getMain()` utility function to first query for an  `entryFunction` metadata entry.
If such metadata is available, the explicitly provided node is used instead of trying to find the main function by name. 
This is particularly useful for Fortran, since there is no notion of a special `main` function.

The following is an example call graph using the new `entryFunction` metadata:
```
{
  "_CG": {
    "meta": {
      "entryFunction": "0"
    },
    "nodes": {
      "0": {
        "callees": null,
        "functionName": "thisIsTheMainFunction",
        "hasBody": true,
        "meta": {},
        "origin": "main.cpp"
      }
    }
  },
  "_MetaCG": {
    "generator": {
      "name": "MetaCG",
      "sha": "4ad73ebf7a108aa388d232ee4824bada13feaa4c",
      "version": "0.8"
    },
    "version": "4.0"
  }
}

``` 

Alongside the implementation, this patch includes both unit and integration tests.

### DIFF
--- a/graph/src/Callgraph.cpp
+++ b/graph/src/Callgraph.cpp
@@ -71,6 +71,15 @@ bool Callgraph::erase(NodeId id) {
   // Check if this is the cached main function
   if (mainNode == ptr.get()) {
     mainNode = nullptr;
+    // Warn if the entry function metadata is still attached.
+    if (auto md = get<EntryFunctionMD>(); md && md->getEntryFunctionId() == id) {
+      MCGLogger::logWarn(
+          "The previous entry function '{}' was erased, but the global metadata entry still points to it."
+          "This metadata will be removed to maintain consistency. Remove or modfity the metadata explicitly to quiet "
+          "this warning.",
+          ptr->getFunctionName());
+      erase<EntryFunctionMD>();
+    }
   }
   std::string name = ptr->getFunctionName();
   nameIdMap[name].erase(std::find(nameIdMap[name].begin(), nameIdMap[name].end(), id));
--- a/graph/test/unit/MCGManagerTest.cpp
+++ b/graph/test/unit/MCGManagerTest.cpp
@@ -365,3 +365,16 @@ TEST_F(MCGManagerTest, GetMainTest) {
   detectedMain = cg->getMain(true);
   ASSERT_EQ(detectedMain->getId(), realMain.getId());
 }
+
+TEST_F(MCGManagerTest, EraseMainTest) {
+  auto& mcgm = metacg::graph::MCGManager::get();
+  mcgm.addToManagedGraphs("newCG", std::make_unique<metacg::Callgraph>(), true);
+
+  auto cg = mcgm.getCallgraph();
+  auto& main = cg->getOrInsertNode("thisIsMain");
+  cg->getOrCreate<metacg::EntryFunctionMD>(main);
+  ASSERT_EQ(cg->getMain(), &main);
+  cg->erase(main.id);
+  ASSERT_EQ(cg->getMain(), nullptr);
+  ASSERT_FALSE(cg->has<metacg::EntryFunctionMD>());
+}


graph/include/CgNode.h
```diff
@@ -202,30 +105,14 @@ class CgNode {
...
-
  public:
   const NodeId id;
 
  private:
   std::string functionName;
   std::optional<std::string> origin;
   bool hasBody;
-  std::unordered_map<std::string, std::unique_ptr<MetaData>> metaFields;
+  //  std::unordered_map<std::string, std::unique_ptr<MetaData>> metaFields;
```
pearzt, line 115: Leftover

graph/src/Callgraph.cpp
```diff
@@ -14,11 +15,22 @@ int metacg_RegistryInstanceCounter{0};
...
+  if (forceRecompute) {
+    mainNode = nullptr;
+  } else if (mainNode) {
     return mainNode;
   }
 
+  // First, check if there is "entryFunction" metadata.
+  if (const auto* md = get<EntryFunctionMD>(); md) {
+    auto id = md->getEntryFunctionId();
+    if (id && (mainNode = getNode(*id))) {
```
jplehr, line 28: Is `id` of pointer type?

graph/src/Callgraph.cpp
```diff
@@ -56,6 +68,10 @@ bool Callgraph::erase(NodeId id) {
   // Destroy the node
   auto& ptr = nodes[id];
   assert(ptr && "The ID must correspond to a valid node");
+  // Check if this is the cached main function
+  if (mainNode == ptr.get()) {
+    mainNode = nullptr;
```
jplehr, line 73: Does this also need to reset the `EntryFunctionMD` metadata?

graph/src/io/VersionFourMCGReader.cpp
```diff
@@ -132,13 +132,19 @@ std::unique_ptr<metacg::Callgraph> metacg::io::VersionFourMetaCGReader::read() {
 
   auto cg = std::make_unique<Callgraph>();
 
+  auto& jNodes = jsonCG["nodes"];
```
jplehr, line 135: This makes "this version of file format v4" incompatible with "the prior version of file format v4".

jplehr (changes requested):
I think the changes are a good direction, but I disagree that this is file format v4.
This is a breaking change.

So, I would prefer to bump the version to v5 instead. What does a versioned file format help us, if we can have two files in a particular format version that are incompatible with each other / the reader that says "I can handle this version of the file format"?

graph/include/metadata/MetadataMixin.h
```diff
@@ -0,0 +1,154 @@
...
+
+  // Delete copy constructor and assignment.
+  MetadataMixin(const MetadataMixin&) = delete;
+  MetadataMixin& operator=(const MetadataMixin&) = delete;
+
+  // Support moving
+  MetadataMixin(MetadataMixin&&) noexcept = default;
+  MetadataMixin& operator=(MetadataMixin&&) noexcept = default;
+
+  ~MetadataMixin() = default;
```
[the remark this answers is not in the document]
sebastiankreutzer: I left it non-virtual on purpose, because the mixin class is not supposed to be used for dynamic polymorphism. 
I.e. you would never call something like:
```
MetadataMixin* base = new Callgraph();
delete base;
```

jplehr:
The reasoning for it being that in a perfect world, a user hands a file to the library and the library knows how to parse it. That is not the case with such changes, because now it depends on the point in time you built the library.

I completely understand the question about why have a `devel` branch then and I don't think you can always foresee what you need. So, yes, you need a moving target.

The motivation with the file format versions was initially more along the lines of something rather internal. So, we can bump the file format version as often as we want. This may / will lead to jumps in the file format version between releases, e.g., going from 3 to 6 or whatever. This is motivated with some painful experiences in "the early days" where things simply broke because of the evolving format and one could no longer reproduce / use the tools.
That likely has evolved into something more "standardese" in nature, or at least is perceived as such, and hence only needs to be "final" or "stable" for a release -- which is different from the original motivation. Not wrong of course, but different.
Given that it is used across tools as a command line argument is not necessarily by design, but due to the fact that historically not all components were based on the graph lib and so it needed to be specified explicitly. That probably added to the perception of it being a "public definition".

In any case, if the motivation of the file format version being something rather internal is perceived as wrong and it should instead be something that is only meaningful in releases, then OK. I won't oppose that direction.
I still want to point out the potential annoyances this can bring with files generated between such changes on `devel` that may simply not be apparent initially.

jplehr:
We agreed in an offline discussion that going forward, the file format will only be considered "fix" in release versions.

pearzt (approved):

graph/src/Callgraph.cpp
jplehr: Ah ok, it's in an optional. Thank you.

PR #71, opened 2025-08-07
[NFC] Add common Spack environment paths to .gitignore

pearzt:
Ignores paths that come with using a Spack environment to describe the development setup. Additionally, I removed the trailing slash after `build*` to also ignore symlinks to build directories, which is something that can happen when using `spack develop`.


jplehr (approved):
LGTM

PR #73, opened 2025-08-07
Don't install external nlohmann-json

pearzt:
This PR proposes two changes: When `METACG_USE_EXTERNAL_JSON` is `ON`, it no longer tries to install the library along with MetaCG. Also, it adds a call to `find_dependency(...)` to `metacgConfig.cmake` if we use an external nlohmann-json, so that projects that use MetaCG are pointed towards the json library that MetaCG was build with and no longer need to provide it themselves.

### DIFF
--- a/cmake/install-config.cmake.in
+++ b/cmake/install-config.cmake.in
@@ -2,7 +2,7 @@ include(CMakeFindDependencyMacro)
 
 find_dependency(Threads)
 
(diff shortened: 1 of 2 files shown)


pearzt:
@TimHeldmann The description was missing still, that's why it was a draft. No, it's not related to the CI regression, but I would like to merge that fix before I merge any of the other PRs. This is now ready for your review, though!

jplehr (approved):

PR #77, opened 2025-08-08
Use fmt::vformat in LoggerUtil to fix problem in new spdlog with C++23

pearzt:
This PR fixes a problem that occurs with spdlog when enabling C++23 (as is needed for CaPI). Internally, `fmt` then switches to prefering constexpr strings, but offers `fmt::vformat` for dynamic strings. With this PR, we propose to switch to `fmt::vformat` in `LoggerUtil.h`. Also, we add a simple logging statement to `graph/test/install/main.cpp` to test using the logger from a project that uses MetaCG.

Officially, we don't yet support anything other than GCC 11, so the fix is currently untested by our CI. Still, this fix will make our life easier as soon as we start to support GCC >= 13.


jplehr (approved):
LGTM

PR #82, opened 2025-08-22
Add CaGe LLVM call graph generator

sebastiankreutzer:
This patch adds CaGe, the call graph generator plugin for LLVM developed by @TimHeldmann.

This version is slightly modified from the original project to fit in with the rest of the MetaCG project.
So far, this patch only contains functionality to register the pass and build the call graph.
It does not include any call graph embedding functionality or runtime library for now.

To make this project easily extendable for different use cases, the pass can register different `CallGraphConsumer`s.
At the moment, the only available consumer is the `FileExporter`, which writes the call graph to a file.

I adopted the integration tests from CGC2, removing those that are either incompatible or irrelevant. 
Tests can be run by executing `./runCaGeTests.sh` in `<build_dir>/tools/cage/test/`.


pearzt (approved):
LGTM. (I did not review the tests.)

tools/README.md
```diff
@@ -2,6 +2,49 @@
 
 This folder contains tools for creating and merging call graphs using the MetaCG graph library.
 
+## CaGe
+
+CaGe is MetaCG's link-time call graph generator.
+To use it, the target applications needs to be built with (full) LTO using the LLD linker.
```
silas-martens: Typo

tools/cage/src/generator/CallGraphGenerator.cpp
```diff
@@ -0,0 +1,170 @@
...
+    // TODO: Improve this design if we want to support multiple virtual call resolution mechanisms, e.g. with
+    //       template policy parameter.
+#ifdef HAVE_METAVIRT
+    auto vcallData = metavirt::vcall_data_for(&I);
+    if (!vcallData.has_value())
+      return 0;
+    if (vcallData.value().call_targets.empty())
+      return 0;
+
+    outs() << metavirt::fn_names_and_origins(vcallData.value()).size() << "\n";
```
silas-martens: is this a debug leftover?

tools/README.md
silas-martens:
```suggestion
To use it, the target application needs to be built with (full) LTO using the LLD linker.
```

tools/cage/src/generator/CallGraphGenerator.cpp
sebastiankreutzer: Yep!

PR #83, opened 2025-08-22
Add metacg-config tool

sebastiankreutzer:
This patch adds `metacg-config`, a lightweight tool inspired by `scorep-config` that can be used to print relevant information about the package.
As a start, the options `--version`, `--revision` and `--prefix` are supported.

The main reason I added this was to make integration of compiler passes (e.g. from cgpatch and later CaGe) easier for the user.
For example, to compile with cgpatch, one could use: `clang++ $(metacg-config --cgpatch-cxxflags) input.cpp`


.github/workflows/mcg-ci.yml
```diff
@@ -45,19 +45,25 @@ jobs:
...
     - uses: actions/checkout@v4
     - name: cmake
-      run: cmake -DCMAKE_POLICY_VERSION_MINIMUM=3.5 -B build -S .
+      run: cmake -DCMAKE_POLICY_VERSION_MINIMUM=3.5 -DCMAKE_INSTALL_PREFIX="$(pwd)/install" -B build -S .
     - name: build
       run: cmake --build build --parallel
     - name: install
       run: |
-        cmake --install build --prefix install
+        cmake  --install build --prefix install
```
jplehr: Nit

config/MCGConfig.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+  if (result.contains("revision")) {
+    std::cout << MetaCG_GIT_SHA;
+    return EXIT_SUCCESS;
+  }
+
+  if (result.contains("version")) {
+    std::cout << MetaCG_VERSION_MAJOR << "." << MetaCG_VERSION_MINOR;
+    return EXIT_SUCCESS;
+  }
```
jplehr: This is covered earlier in lines 34-37

config/MCGConfig.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+  // Exactly one of these is allowed at the same time
+  int optCount = result.count("version") + result.count("revision") + result.count("prefix");
+  if (optCount == 0) {
+    std::cerr << "Error: No command specified.\n";
+    return EXIT_FAILURE;
+  } else if (optCount > 1) {
+    std::cerr << "Warning: Multiple mutually exclusive commands specified. Only one of them will be processed.\n";
```
jplehr: Should this `EXIT_FAILURE` instead?
I have no strong feelings about this.

config/MCGConfig.cpp
```diff
@@ -0,0 +1,55 @@
...
+    return EXIT_SUCCESS;
+  }
+
+  if (result.contains("prefix")) {
+    std::cout << INSTALL_PREFIX;
+    return EXIT_SUCCESS;
+  }
+
+  return EXIT_SUCCESS;
+}
```
jplehr: EOF missing?

jplehr (commented):
I think this should also go into the existing `tools` folder instead.

sebastiankreutzer:
> I think this should also go into the existing `tools` folder instead.

My rationale was that it is not a "tool" like the other existing tools, as it does not use the graph lib in any way.
Also, I think it should not be tied to the `METACG_BUILD_GRAPH_TOOLS` option, as it is useful even if none of the tools are built.

config/MCGConfig.cpp
sebastiankreutzer: I was copying the behavior of `scorep-config` here, which silently ignores superfluous commands. I'm happy to change this though.

sebastiankreutzer:
Maybe a `util` folder would be a suitable alternative?

config/MCGConfig.cpp
jplehr: Well, if there is precedent in other such `xyz-config` tools then I would believe we should not deviate.

jplehr:
> > I think this should also go into the existing `tools` folder instead.
> 
> My rationale was that it is not a "tool" like the other existing tools, as it does not use the graph lib in any way. Also, I think it should not be tied to the `METACG_BUILD_GRAPH_TOOLS` option, as it is useful even if none of the tools are built.

Understood. I mostly want to prevent that we start littering the top-level with stuff for everything that is "not quite the same" as to what is already in tree. Do we have other things that are planned to come into MetaCG that are mostly utils?
The CMake target collector util might be another of such utils. That would warrant a TLD `utils` I suppose.

sebastiankreutzer:
> > > I think this should also go into the existing `tools` folder instead.
> > 
> > 
> > My rationale was that it is not a "tool" like the other existing tools, as it does not use the graph lib in any way. Also, I think it should not be tied to the `METACG_BUILD_GRAPH_TOOLS` option, as it is useful even if none of the tools are built.
> 
> Understood. I mostly want to prevent that we start littering the top-level with stuff for everything that is "not quite the same" as to what is already in tree. Do we have other things that are planned to come into MetaCG that are mostly utils? The CMake target collector util might be another of such utils. That would warrant a TLD `utils` I suppose.

We just had a brief discussion about this with @TimHeldmann and @pearzt. The consensus was that:
- It should not go into `tools`, as it is a project-scope utility, not a graphlib one.
- There are currently no other utilities of this kind, so adding an additional `utils` folder is a bit premature. If we change our mind or other such utilities are added, we can easily move it.

pearzt (approved):

jplehr:
> > > > I think this should also go into the existing `tools` folder instead.
> > > 
> > > 
> > > My rationale was that it is not a "tool" like the other existing tools, as it does not use the graph lib in any way. Also, I think it should not be tied to the `METACG_BUILD_GRAPH_TOOLS` option, as it is useful even if none of the tools are built.
> > 
> > 
> > Understood. I mostly want to prevent that we start littering the top-level with stuff for everything that is "not quite the same" as to what is already in tree. Do we have other things that are planned to come into MetaCG that are mostly utils? The CMake target collector util might be another of such utils. That would warrant a TLD `utils` I suppose.
> 
> We just had a brief discussion about this with @TimHeldmann and @pearzt. The consensus was that:
> 
> * It should not go into `tools`, as it is a project-scope utility, not a graphlib one.
> * There are currently no other utilities of this kind, so adding an additional `utils` folder is a bit premature. If we change our mind or other such utilities are added, we can easily move it.

What about the `TargetCollector.py`? Isn't this a util?

sebastiankreutzer:
> > > > > I think this should also go into the existing `tools` folder instead.
> > > > 
> > > > 
> > > > My rationale was that it is not a "tool" like the other existing tools, as it does not use the graph lib in any way. Also, I think it should not be tied to the `METACG_BUILD_GRAPH_TOOLS` option, as it is useful even if none of the tools are built.
> > > 
> > > 
> > > Understood. I mostly want to prevent that we start littering the top-level with stuff for everything that is "not quite the same" as to what is already in tree. Do we have other things that are planned to come into MetaCG that are mostly utils? The CMake target collector util might be another of such utils. That would warrant a TLD `utils` I suppose.
> > 
> > 
> > We just had a brief discussion about this with @TimHeldmann and @pearzt. The consensus was that:
> > 
> > * It should not go into `tools`, as it is a project-scope utility, not a graphlib one.
> > * There are currently no other utilities of this kind, so adding an additional `utils` folder is a bit premature. If we change our mind or other such utilities are added, we can easily move it.
> 
> What about the `TargetCollector.py`? Isn't this a util?

I guess I considered `TargetCollector.py` a `cgcollector` specific feature that should really have been placed into `cgcollector`.
But looking at it now, it should also work with the graph tools. 
So I'm happy to move into a common `utils` folder.

jplehr:
> > > > > > I think this should also go into the existing `tools` folder instead.
> > > > > 
> > > > > 
> > > > > My rationale was that it is not a "tool" like the other existing tools, as it does not use the graph lib in any way. Also, I think it should not be tied to the `METACG_BUILD_GRAPH_TOOLS` option, as it is useful even if none of the tools are built.
> > > > 
> > > > 
> > > > Understood. I mostly want to prevent that we start littering the top-level with stuff for everything that is "not quite the same" as to what is already in tree. Do we have other things that are planned to come into MetaCG that are mostly utils? The CMake target collector util might be another of such utils. That would warrant a TLD `utils` I suppose.
> > > 
> > > 
> > > We just had a brief discussion about this with @TimHeldmann and @pearzt. The consensus was that:
> > > 
> > > * It should not go into `tools`, as it is a project-scope utility, not a graphlib one.
> > > * There are currently no other utilities of this kind, so adding an additional `utils` folder is a bit premature. If we change our mind or other such utilities are added, we can easily move it.
> > 
> > 
> > What about the `TargetCollector.py`? Isn't this a util?
> 
> I guess I considered `TargetCollector.py` a `cgcollector` specific feature that should really have been placed into `cgcollector`. But looking at it now, it should also work with the graph tools. So I'm happy to move into a common `utils` folder.

I personally would treat both of those tools as utils.

PR #86, opened 2025-09-04
Handle inconsistent isVirtual entry when reading V2 CGs

sebastiankreutzer:
As it turns out, `CGCollector` sometimes emits `isVirtual: false` even if the function is actually virtual and there are entries in `overrides` and `overriddenBy`. 
This leads to inconsistent internal override metadata, because this metadata is only attached if `isVirtual` is true.

This patch fixes the issue on the reader site, by not relying on `isVirtual` to determine if there are overriding or overridden functions.
A warning is printed to alert the user of inconsistent entries.

### DIFF
--- a/graph/src/io/VersionTwoMCGReader.cpp
+++ b/graph/src/io/VersionTwoMCGReader.cpp
@@ -97,8 +97,12 @@ void metacg::io::VersionTwoMCGReader::upgradeV2FormatToV4Format(nlohmann::json& 
       jNode["origin"] = nullptr;
     }
 
+    bool definesOverrideData = !jNode.at("overrides").empty() || !jNode.at("overriddenBy").empty();
+    if (!jNode.at("isVirtual").get<bool>() && definesOverrideData) {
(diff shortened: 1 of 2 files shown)


pearzt (approved):
LGTM

jplehr:
Do we know and care about when that happens?
Edit: Ultimate question then is whether it is fixed with CGC2?

TimHeldmann:
This apparently also includes a broken line ending.

TimHeldmann:
> Edit: Ultimate question then is whether it is fixed with CGC2?

CGC writes its own json.
As CGC2 uses the graph lib to emit its json, if the graph lib is correct, CGC2 is as well

jplehr:
> > Edit: Ultimate question then is whether it is fixed with CGC2?
> 
> 
> 
> CGC writes its own json.
> 
> As CGC2 uses the graph lib to emit its json, if the graph lib is correct, CGC2 is as well

Yes, but the problem, at least as I understand the PR description arises because something in CGC isn't doing the right thing in the first place. So I don't know if that problem also is present in CGC2. 
Edit: or is this really only a problem of the emitting process and the information inside the collector is actually correct?

PR #87, opened 2025-09-08
Fix bug in Callgraph::erase where edge caches were only partially updated

pearzt:
The caching mechanism is something that we should consider reworking at some point, since it was not designed with node deletion in mind. For now, though, this fix makes sure to update the caller/callee cache also for nodes that are adjacent to the erased node.


jplehr (approved):
LGTM

PR #88, opened 2025-09-13
[CGC2] Fix error in test script

jplehr:
This error leads to single-file test errors being ignored as observed in a local run of the tests.


jplehr:
The error in the workflow is exposing an issue that was there before. 
Not related to this PR but we need to fix it.

sebastiankreutzer (approved):

TimHeldmann (approved):
Looks good to me.
We should merge this, even if it turns the pipeline red.

PR #89, opened 2025-09-13
[CI] Fix build path for cgc2 testing downstream

jplehr:
In the downstream CI the build directory name depends on the different compiler modules that are loaded, hence we need to make it explicit.


pearzt (approved):
LGTM

PR #91, opened 2025-09-20
[Graph] Fix member visibility to private

jplehr:


sebastiankreutzer (approved):
IMO exposing `id` publicly is fine because it is `const`. However, it's probably a good idea to provide either public access or a getter method, and not both, as it was previously implemented. So I'm happy with this change.

PR #94, opened 2025-10-22
[NFC] Fix logic to get source file name when using LLVM 21

pearzt:
We don't support LLVM 21 yet, but as soon as we do, we will need this. (Also, I need MetaCG with LLVM 21 right now, so that's why there's is a fix.)


jplehr (approved):
lgtm

PR #96, opened 2025-10-31
Allow reading null CG in v2

silas-martens:
This patch enables reading call graphs with `CG: null` entry .

### DIFF
--- a/graph/src/io/VersionTwoMCGReader.cpp
+++ b/graph/src/io/VersionTwoMCGReader.cpp
@@ -51,6 +51,7 @@ std::unique_ptr<metacg::Callgraph> metacg::io::VersionTwoMCGReader::read() {
 
   MCGFileInfo fileInfo{ffInfo, genVersionInfo};
   if (j.at(ffInfo.cgFieldName).is_null()) {
+    j[ffInfo.cgFieldName] = nlohmann::json::object();
     const std::string warningMsg = "Detected null call graph in metacg file; proceeding with an empty call graph.";
     MCGLogger::logWarn(warningMsg);
   }
--- a/graph/src/io/VersionTwoMCGReader.cpp
+++ b/graph/src/io/VersionTwoMCGReader.cpp
@@ -50,7 +50,12 @@ std::unique_ptr<metacg::Callgraph> metacg::io::VersionTwoMCGReader::read() {
                 generatorVersion);
 
   MCGFileInfo fileInfo{ffInfo, genVersionInfo};
-  if (!j.contains(ffInfo.cgFieldName) || j.at(ffInfo.cgFieldName).is_null()) {
+  if (j.at(ffInfo.cgFieldName).is_null()) {
+    const std::string warningMsg = "Detected null call graph in metacg file; proceeding with an empty call graph.";
+    MCGLogger::logWarn(warningMsg);
+  }
+
+  if (!j.contains(ffInfo.cgFieldName)) {
     const std::string errorMsg = "The call graph in the metacg file was not found or null.";
     errConsole->error(errorMsg);
     throw std::runtime_error(errorMsg);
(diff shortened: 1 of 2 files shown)


jplehr:
Does this work because we auto-upgrade to v4?

pearzt (approved):

pearzt:
@jplehr I'm not sure I understand your question. We only have one in-memory representation that can be constructed from both file format versions. So, if you consider the in-memory representation as v4 (as that's the one it has been designed for), then yes, we "auto-upgrade" to v4.

jplehr:
I guess my expectation was that we would need code changes beyond removing the condition to bail out.

edit:
I just looked at the code again, and it seems we do an auto-promo to v4 and then use the v4 reader, cf., lines 66-67 (or so).

jplehr:
I assume that this is only in-memory and does not lift the original input file or at least not in-place?
I think that would be an unexpected side effect of this change.

silas-martens:
> I guess my expectation was that we would need code changes beyond removing the condition to bail out.
> 
> edit: I just looked at the code again, and it seems we do an auto-promo to v4 and then use the v4 reader, cf., lines 66-67 (or so).

This fix worked in cgdiff and pipeline was not failing so I just assumed I dont break anything :). I made a bit more research and came to the following conclusion:

If the `_CG` field is null, we call `upgradeV2FormatToV4Format(null)`. In this function, we then iterate over `null.items()` (`for (auto& it : j.items())`). One would naturally assume this would fail, but `j.items()` internally calls `j::begin()`, and `begin()` has a no-throw guarantee. Therefore, the loop simply behaves as if j were an empty JSON object.

To make this clearer, I added `j[cgFieldName] = json::object();` when it is null.

jplehr:
> > I guess my expectation was that we would need code changes beyond removing the condition to bail out.
> 
> > 
> 
> > edit: I just looked at the code again, and it seems we do an auto-promo to v4 and then use the v4 reader, cf., lines 66-67 (or so).
> 
> 
> 
> This fix worked in cgdiff and pipeline was not failing so I just assumed I dont break anything :). I made a bit more research and came to the following conclusion:
> 
> 
> 
> If the `_CG` field is null, we call `upgradeV2FormatToV4Format(null)`. In this function, we then iterate over `null.items()` (`for (auto& it : j.items())`). One would naturally assume this would fail, but `j.items()` internally calls `j::begin()`, and `begin()` has a no-throw guarantee. Therefore, the loop simply behaves as if j were an empty JSON object.
> 
> 
> 
> To make this clearer, I added `j[cgFieldName] = json::object();` when it is null. 

Thank you for that walk through. I am fine with the implicit stuff. I only want to understand if the promotion changes the actual file on disk.
I think it would be surprising to have the file on disk change when it was not requested.

silas-martens:
> > > I guess my expectation was that we would need code changes beyond removing the condition to bail out.
> > 
> > 
> > > 
> > 
> > 
> > > edit: I just looked at the code again, and it seems we do an auto-promo to v4 and then use the v4 reader, cf., lines 66-67 (or so).
> > 
> > 
> > This fix worked in cgdiff and pipeline was not failing so I just assumed I dont break anything :). I made a bit more research and came to the following conclusion:
> > If the `_CG` field is null, we call `upgradeV2FormatToV4Format(null)`. In this function, we then iterate over `null.items()` (`for (auto& it : j.items())`). One would naturally assume this would fail, but `j.items()` internally calls `j::begin()`, and `begin()` has a no-throw guarantee. Therefore, the loop simply behaves as if j were an empty JSON object.
> > To make this clearer, I added `j[cgFieldName] = json::object();` when it is null.
> 
> Thank you for that walk through. I am fine with the implicit stuff. I only want to understand if the promotion changes the actual file on disk. I think it would be surprising to have the file on disk change when it was not requested.

The upgradeV2FormatToV4Format only modifies the JSON in memory. It does not write anything back to disk as far as I understand.

jplehr (approved):

PR #98, opened 2025-11-14
[CMake] Add cache files for convenience

jplehr:
Adding two CMake cache files to make building MetaCG more convenient.

One cache is a full build that requires the external dependencies to be installed via the build_submodules.sh script.

The other cache is a build of the "new" graph tools that does not depend on the external components.


jplehr:
Disclaimer: I did not test the full build file. :)

PR #101, opened 2025-11-26
Fix test failures for CGC2 

TimHeldmann:
To have all tests complete for CGC2 we need to

1. Fix the C++ standard to have a common behavior between versions 
2. Disables system configuration dependend tests
3. Relax the strictness regarding caller callee observations:
 We currently abort execution if we try to add an edge between to functions if the called function has not been seen yet.
 We already relaxed this for lambdas and default constructors/destructors (refer to my rant about compiler generated code not having to adhere to declare-use order). With this we relaxed this again, as this seems to also happen for template instantiated code. 
 
 This patch adds these three fixes and is validated by the down-stream CI.
 
  Looking forward to your feedback.


tools/cgcollector2/src/CallGraphCollectionAction.cpp
```diff
@@ -67,18 +67,21 @@ void CallGraphCollectorConsumer::HandleTranslationUnit(clang::ASTContext& Contex
     case 1: SPDLOG_WARN("This tool can not generate output files in the V1 format, using V2 instead");
       __attribute__ ((fallthrough));
     case 2:
-      mcgWriter = std::make_unique<metacg::io::VersionTwoMCGWriter>();
+      mcgWriter = std::make_unique<metacg::io::VersionTwoMCGWriter>(metacg::getVersionTwoFileInfo({std::string("CGCollector2"), MetaCG_VERSION_MAJOR,
```
jplehr: Why is this change needed?

tools/cgcollector2/src/CallGraphNodeGenerator.cpp
```diff
@@ -167,7 +168,7 @@ bool CallGraphNodeGenerator::shouldIncludeFunction(const Decl* D) {
     }
 
     // if it is a constructor or destructor, and we don't capture those, return early
-    if ((isa<CXXConstructorDecl>(D) || isa<CXXDestructorDecl>(D)) && !captureCtorsDtors) {
+    if ((isa<CXXConstructorDecl>(D) || isa<CXXDestructorDecl>(D)) && !captureCtorsDtors ) {
```
jplehr: format issue

tools/cgcollector2/src/CallGraphNodeGenerator.cpp
```diff
@@ -462,6 +447,16 @@ void CallGraphNodeGenerator::addEdge(clang::Decl* Child) {
   for (auto& parentName : getMangledNames(topLevelFD)) {
     for (auto& childName : getMangledNames(clang::cast<clang::NamedDecl>(Child))) {
       assert(callgraph->hasNode(parentName));
+      if(!callgraph->hasNode(childName)){
+        assert(isa<FunctionDecl>(Child));
+        //This can happen if we have an immediately invoked lambda,
+        // or infer a call to a destructor via a param variable declaration not referring to a definition
+        // FIXME: and other cases we do not understand.
```
jplehr: Do we have a (better) understanding of this FIXME now?

tools/cgcollector2/test/input/cxxRecordCalls/0008.cpp
```diff
@@ -1,15 +1,13 @@
 /**
-* File: GlobalStackClass.cpp
+* File: GlobalStackStruct.cpp
 * License: Part of the MetaCG project. Licensed under BSD 3 clause license. See LICENSE.txt file at
 * https://github.com/tudasc/metacg/LICENSE.txt
 */
 
-class localClass {
- public:
+struct localStruct {
```
jplehr, line 7: Why this change?

tools/cgcollector2/test/input/cxxRecordCalls/0005.cpp_disabled
```diff
@@ -0,0 +1,15 @@
+/**
+* File: GlobalLambda.cpp
+* License: Part of the MetaCG project. Licensed under BSD 3 clause license. See LICENSE.txt file at
+* https://github.com/tudasc/metacg/LICENSE.txt
+*/
+
+//This test is currently disabled, as lambda parameter lists contain a full path to the definition
```
jplehr: Can we add a TODO / FIXME here please.

```suggestion
// TODO: This test is currently disabled, as lambda parameter lists contain a full path to the definition
```

jplehr (commented):
Why are so many tests changed in this PR?
At least to me, some/most of these changes seem unnecessary.

tools/cgcollector2/src/CallGraphCollectionAction.cpp
TimHeldmann: Because the generator Info default is: File was generated by MetaCG.
Files coming out of CGC2 are generated by CGC2, and this change reflects this.

tools/cgcollector2/src/CallGraphNodeGenerator.cpp
TimHeldmann: Fixed

tools/cgcollector2/src/CallGraphNodeGenerator.cpp
TimHeldmann: No we don't.
This Fixme is new.
We had two occurences of the relaxed node checking already.
Found a third one (the template instantiation thing I mentioned in the PR description) and I added this Fixme.
I could not find a regularity in the behavior, but someone else might have the time and the understanding to find out why this happens

pearzt (approved):
Just discussed all non-test changes with @TimHeldmann, they all LGTM. Have only reviewed test changes on random sample basis.

As soon as @jplehr's feedback is addressed, I think this is ready to merge.

tools/cgcollector2/test/input/cxxRecordCalls/0005.cpp_disabled
TimHeldmann: Added TODO

tools/cgcollector2/test/input/cxxRecordCalls/0008.cpp
TimHeldmann: This is not really a change.
Git just got a little confused.

Test cxxRecordCalls/0001.cpp and cxxRecordCalls/0002.cpp got moved to the FunctionPointer and SingleTU tests respectively, as they do not pertain to cxxRecords.
This means we needed to rename the files so all tests moved down two places, so our test script still works, as it depends on the numbering scheme.
Previously this test was named 0010.cpp, the contents stayed identical.

Git now thinks that all the test files got changed, even if most of them only got renamed.

TimHeldmann:
> Why are so many tests changed in this PR? At least to me, some/most of these changes seem unnecessary.

Most of the tests did not change, but git got confused due to renaming some of them.
I tried my best to explain it [here](https://github.com/tudasc/MetaCG/pull/101#discussion_r2564292105)

jplehr (approved):

PR #103, opened 2025-11-28
Replace old testers with CGDiff

silas-martens:
I replaced some of the old testers with CGDiff. 

For CGC2 and CGPatch this worked well. For CGCollector, I processed the diff outputs to ensure that all edges in the generated call graph are contained in the ground truth, replicating the behavior of the previous comparison.

I also made a small fix to CGDiff to compare metadata values order-independently.
As discussed during the hackathon, I removed the AA and format-one tests.


cgcollector/test/testBase.sh
```diff
@@ -1,50 +1,72 @@
 
 cgcollectorExe=cgcollector
-testerExe=cgsimpletester
 cgmergeExe=cgmerge
 build_dir=build # may be changed with opt 'b'
+diffFile=temp.json
```
sebastiankreutzer: Instead of a hardcoded filename, use `mktemp`, e.g.: `diffFile=$(mktemp temp.json.XXX)`

tools/cgdiff/CGDiff.cpp
```diff
@@ -66,8 +66,17 @@ std::vector<std::unique_ptr<metacg::Diff>> compare(const metacg::Callgraph& mcgA
...
 
+          if (json.is_object()) {
+            for (auto& [subKey, subValue] : json.items()) {
+              if (subValue.is_array()) {
+                std::sort(subValue.begin(), subValue.end());
+              }
+            }
+          }
+          std::string value = json.dump(-1);
+          std::cout << "Key: " << key << " value: " << value << std::endl;
```
sebastiankreutzer: Left over from debugging?

tools/cgdiff/CGDiff.cpp
```diff
@@ -150,7 +159,7 @@ int main(int argc, char** argv) {
                              "Returns 0 if call-graphs are equal, 1 otherwise.\n");
 
     options.add_options()("ignore-edges", "ignore edges", cxxopts::value<bool>()->default_value("false"))(
-        "ignore-hasBody", "ignore hasBody", cxxopts::value<bool>()->default_value("false"))(
+      "ignore-hasBody", "ignore hasBody", cxxopts::value<bool>()->default_value("false"))(
```
sebastiankreutzer: Indentation correct here?

sebastiankreutzer (commented):
Great work! I added a few minor comments here and there.

.gitlab-ci.yml
```diff
@@ -168,7 +168,6 @@ test-cgc:
     - module load clang/$LLVM
     - cd cgcollector/test
     - mkdir -p log
-    - bash run_format_one_test.sh -b $MCG_BUILD
```
TimHeldmann, line 171: Why were the format one tests disabled?
I think I remember discussing this, but I don't remember what we agreed on.

cgcollector/test/run_format_two_test.sh
```diff
@@ -2,16 +2,15 @@
...
-#if [ command -v $testerExe ]; then
 if [[ $(type -P $testerExe) ]]; then
-  echo "The MCGTester binary (mcgtester) could not be found in path, testing with relative path."
+  echo "The cgdiff binary (cgdiff) could not be found in path, testing with relative path."
 fi
-stat ../../${build_dir}/cgcollector/test/mcgtester >>log/testrun.log 2>&1
+stat ../../${build_dir}/tools/cgdiff/cgdiff >> log/testrun.log 2>&1
 if [ $? -eq 1 ]; then
-  echo "The file seems also non-present in ../${build_dir}/test. Aborting test. Failure! Please build the tester first."
+  echo "The file seems also non-present in ../${build_dir}/cgdiff/. Aborting test. Failure! Please build the tester first."
```
TimHeldmann: Technically it should now be: "[...]! Please build cgdiff/the difference dettection tool first"

cgcollector/test/testBase.sh
```diff
@@ -1,50 +1,73 @@
...
+    local diffFile="$1"
+
+    if [ ! -f "$diffFile" ]; then
+        echo "Diff file not found: $diffFile"
+        return 2
+    fi
+
+    # Iterate over all nodes
+    local nodes
+    nodes=$(jq -r '.diff.nodeDiffs | keys[]' "$diffFile") || return 2
```
TimHeldmann, line 23: Are you sure this works with the base and devel containers?
`jq` should only be available on 22.04, but the base container is based on 20.04.

cgcollector/test/testBase.sh
```diff
@@ -1,50 +1,73 @@
...
+
+        # Extract metadataOnlyInA and metadataOnlyInB arrays
+        local aJson bJson
+        aJson=$(jq ".diff.nodeDiffs[\"$node\"].metadataOnlyInA" "$diffFile")
+        bJson=$(jq ".diff.nodeDiffs[\"$node\"].metadataOnlyInB" "$diffFile")
+
+        # Filter out entries that contain numberOfControlFlowOps
+        local aFiltered bFiltered
+        aFiltered=$(echo "$aJson" | jq '[.[] | select(test("numberOfControlFlowOps") | not)]')
+        bFiltered=$(echo "$bJson" | jq '[.[] | select(test("numberOfControlFlowOps") | not)]')
```
TimHeldmann: Why do we filter?

cgcollector/test/testBase.sh
```diff
@@ -67,102 +90,25 @@ function applyFileFormatTwoToSingleTU {
...
-  $cgcollectorExe --metacg-format-version=2 --capture-ctors-dtors --capture-stack-ctors-dtors --enable-AA ${addFlags} --output ${gfile} $tfile -- >>log/testrun.log 2>&1
-  cat $gfile | python3 -m json.tool > ${gfile}_
-  mv ${gfile}_ ${gfile}
-  $testerExe $tgt $gfile >>log/testrun.log 2>&1
-
-  if [ $? -ne 0 ]; then
-    echo "Failure for file: $gfile. Keeping generated file for inspection"
-    fail=$((fail + 1))
+    if only_metadata_diff $diffFile; then
+        rm $gfile
```
TimHeldmann, line 95: We do not check for metadata equality?
Has this always been the case?

container/full-build
```diff
@@ -14,7 +14,8 @@ RUN apt-get update -y && \
         openmpi-bin openmpi-common bison flex bear \
         lsb-release wget software-properties-common \
         python3-matplotlib python3-pyqt5 ninja-build \
-    libopenmpi-dev libhwloc-dev libevent-dev
+    libopenmpi-dev libhwloc-dev libevent-dev \
+    jq
```
TimHeldmann, line 18: Pretty sure, the other containers should get this treatment as well.

TimHeldmann (commented):
Thank you for your work.
Looking forward to get this merged

cgcollector/test/testBase.sh
silas-martens: The ground truth call-graph is treated as a subset of the generated call-graph. As a result, the generated call-graph may contain additional metadata entries that do not appear in the ground truth. These extra entries are ignored.

In other words, we only verify that the metadata present in the ground truth is also contained in the generated call graph. This replicates the behavior of the previous tool.

cgcollector/test/testBase.sh
silas-martens: The value of numberOfControlFlowOps varies across compiler versions. Therefore, I decided to ignore differences in this metadata value during comparison.

cgcollector/test/testBase.sh
silas-martens: I’ve only tested this in GitHub and GitLab CI pipelines.

.gitlab-ci.yml
silas-martens: They were removed because format one is not supported by the current graph library, and therefore not by cgdiff either. Keeping the tests would have required us to keep the old tester.

As I recall, we leaned towards removing them but never made a final decision. I’ve therefore included the removal in a separate commit: if we decide to keep the tests after all, the change can be reverted easily.

If we do remove these tests, it would make sense to also consider deprecating format one in the near future.

cgcollector/test/testBase.sh
silas-martens: I added jq to base. Are base/devel containers used by anybody?

cgcollector/test/testBase.sh
TimHeldmann: Does this filter the whole key or only the value?
Should we make sure, that this metadata is constructed with a valid integer, even if the values differ?

cgcollector/test/testBase.sh
TimHeldmann: I don't think they are in active use, but as long as they are in the repo, I would not break them on purpose, especially if the fix is relatively easy.

.gitlab-ci.yml
TimHeldmann: Eh, I see.
If I remember correctly CGC is the only tool capable of even writing MetaCG V1 files.
And as CGC might get fazed out in favour of CGC2 I can live with this.

cgcollector/test/testBase.sh
TimHeldmann: Ok, I think it is fine to then replicate this behavior in this PR.
I'll raise an issue, that we should address this in a different PR.

cgcollector/test/testBase.sh
silas-martens: Yeah this part was completely scuffed. I reworked it.
- only_metadata_diff now checks that the metadata is only missing in the grountruth (GT is subset of generated CG)
- before executing cgdiff i set the values of numberOfControlFlowOps to 42 to ignore value differences caused by differing clang versions

Thanks for pointing out

PR #105, opened 2025-12-04
Fix Clang version check for FileEntry::getName()

silas-martens:
`FileEntry::getName()` was deprecated in clang 18 and removed in clang 19 (https://github.com/llvm/llvm-project/commit/6c1dbd5359c4336d03b11faeaea8459b421f2c5c).

This fixes compilation with Clang 20 on my local machine.

### DIFF
--- a/tools/cgcollector2/fileInfoDemoPlugin/FileInfoMetadataPlugin.cpp
+++ b/tools/cgcollector2/fileInfoDemoPlugin/FileInfoMetadataPlugin.cpp
@@ -24,7 +24,7 @@ struct FileInfoMetadataPlugin : Plugin {
     auto& astCtx = functionDecl->getASTContext();
     const auto fullSrcLoc = astCtx.getFullLoc(sourceLocation);
 
-#if LLVM_VERSION_MAJOR < 19
+#if LLVM_VERSION_MAJOR < 18
     const auto fileEntry = fullSrcLoc.getFileEntry();
 #else
     const auto fileEntry = fullSrcLoc.getFileEntryRef();
--- a/tools/cgcollector2/fileInfoDemoPlugin/FileInfoMetadataPlugin.cpp
+++ b/tools/cgcollector2/fileInfoDemoPlugin/FileInfoMetadataPlugin.cpp
@@ -24,7 +24,7 @@ struct FileInfoMetadataPlugin : Plugin {
     auto& astCtx = functionDecl->getASTContext();
     const auto fullSrcLoc = astCtx.getFullLoc(sourceLocation);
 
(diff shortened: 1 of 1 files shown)


tools/cgcollector2/fileInfoDemoPlugin/FileInfoMetadataPlugin.cpp
```diff
@@ -24,7 +24,7 @@ struct FileInfoMetadataPlugin : Plugin {
     auto& astCtx = functionDecl->getASTContext();
     const auto fullSrcLoc = astCtx.getFullLoc(sourceLocation);
 
-#if LLVM_VERSION_MAJOR < 21
+#if LLVM_VERSION_MAJOR < 19
```
TimHeldmann:
```suggestion
#if LLVM_VERSION_MAJOR < 18
```

TimHeldmann (commented):
The earliest point at which `fullSrcLoc.getFileEntryRef()` is availbale is LLVM 18.  
I would prefer to use it as soon as possible, instead of an arbitrary cut-off one version further.

jplehr:
Am I missing something? The commit you linked does not remove the API used in MetaCG.

Reading Tim's comment, I like the idea of using the earliest possible version that has the new API.

pearzt:
@jplehr Although you're right that the commit does not effect `getFileEntry`, Silas did link the correct commit. It removes `getName` from `FileEntry`, so one has to go via `FileEntryRef`: ([relevant lines in commit](https://github.com/llvm/llvm-project/commit/6c1dbd5359c4336d03b11faeaea8459b421f2c5c#diff-62ded32b9a0886b8ad01da1e16a1422d694c9314b174ef1803a54357ff0ccec0L331-L332))

pearzt (approved):
I agree with @TimHeldmann that we should start using `getFileEntryRef` with LLVM 18, otherwise this looks good to me.

silas-martens:
> @jplehr Although you're right that the commit does not effect `getFileEntry`, Silas did link the correct commit. It removes `getName` from `FileEntry`, so one has to go via `FileEntryRef`: ([relevant lines in commit](https://github.com/llvm/llvm-project/commit/6c1dbd5359c4336d03b11faeaea8459b421f2c5c#diff-62ded32b9a0886b8ad01da1e16a1422d694c9314b174ef1803a54357ff0ccec0L331-L332))

Yep, that’s the reason. My compilation with LLVM 20 failed because `FileEntry::getName()` was removed in Clang 19 (linked commit), requiring `FileEntryRef` from LLVM 19 onwards.

PR #107, opened 2025-12-17
Pretty print JSON output of cgdiff

sebastiankreutzer:
Minor change to pretty print the JSON output of `cgdiff`, making it much easier to read.


pearzt (approved):
Maybe we should think about making this configurable in the future, but for now this is definitely an improvement!

PR #108, opened 2025-12-18
[CGQuery] CLI tool for call graph analyses

silas-martens:
**CGQuery** is a command-line tool to perform various analyses on **Metacg** call graphs.

Additional analyses will be added in separate PRs.


tools/cgquery/CGQuery.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+        if (!result.count("source")) {
+            std::cerr << "Please specify the source node" << std::endl;
+            std::cout << options.help() << std::endl;
+            return 2;
+        }
+
+
+
+        auto* sourceNode = cg->getFirstNode(result["source"].as<std::string>());
```
pearzt: Since nodes are not uniquely identified by name, I think we should at the very least output a warning if the name provided by the user is ambiguous.

tools/cgquery/CGQuery.cpp
```diff
@@ -0,0 +1,97 @@
...
+        if (result.count("to")) {
+            std::string toName = result["to"].as<std::string>();
+
+            auto* toNode = cg->getFirstNode(toName);
+            if (!toNode) {
+                std::cerr << "Entry node '" << toName << "' not found in CG.\n";
+                return 1;
+            }
+
+            return reachabilityAnalysis.existsPathBetween(sourceNode, toNode, true) ? 0 : 1;
```
pearzt: Is this valid?
```suggestion
            return !reachabilityAnalysis.existsPathBetween(sourceNode, toNode, true);
```

pearzt (changes requested):
Thank you for tackling this! Looks like a good basis for `cgquery`.

tools/cgquery/CGQuery.cpp
silas-martens: Thanks for the feedback. CGQuery now warns when a function name is not unique and matches multiple nodes. Additionally, CGQuery now interprets strings of integers as node IDs to improve usability with v4 format.

tools/cgquery/CGQuery.cpp
silas-martens: Yes, i dont know how i came up with this :)

pearzt (approved):

PR #110, opened 2026-01-15
Fix race conditions in CGC2 test script

silas-martens:
Previously, `testBase.sh` generated output using fixed filenames, which could lead to race conditions when tests were executed in parallel matrix CI jobs.
This PR resolves this issue by appending a unique identifier to the generated filenames.

### DIFF
--- a/tools/cgcollector2/test/testBase.sh
+++ b/tools/cgcollector2/test/testBase.sh
@@ -3,7 +3,11 @@ testerExe=cgdiff
 cgmergeExe=cgmerge2
 build_dir=build # may be changed with opt 'b'
 
-mkdir -p log
+timeStamp=$(date +%s)
+
+: ${CI_CONCURRENT_ID:="${timeStamp}"}
+LOGDIR="$(pwd)/log_${CI_CONCURRENT_ID}"
+mkdir $LOGDIR
 
 # Function to invoke the CGCollector with file format version 2 to a target source code
 # Param 1: The relative path name to the test case.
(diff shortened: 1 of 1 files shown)


tools/cgcollector2/test/testBase.sh
```diff
@@ -3,7 +3,11 @@ testerExe=cgdiff
 cgmergeExe=cgmerge2
 build_dir=build # may be changed with opt 'b'
 
-mkdir -p log
+timeStamp=$(date +%s)
```
pearzt: We have $CI_CONCURRENT_ID in our downstream CI, but using a timestamp as a fallback to prevent name collisions seems unreliable to me as parallel testing jobs could very well be launched during the same second.

tools/cgcollector2/test/testBase.sh
```diff
@@ -3,7 +3,11 @@ testerExe=cgdiff
...
+
+: ${CI_CONCURRENT_ID:="${timeStamp}"}
+LOGDIR="$(pwd)/log_${CI_CONCURRENT_ID}"
```
pearzt: I think it would be a good idea to keep the test scripts agnostic of the CI infrastructure that executes it and not use environment variables that are only defined in some contexts. How about passing a unique identifier as a parameter or reusing `build_dir` to prevent the collisions?

pearzt (changes requested):
Great, thanks a lot for tackling this!

tools/cgcollector2/test/testBase.sh
jplehr: The `CI_CONCURRENT_ID` env var is set to `timeStamp` when executing outside of the Gitlab runner, so it always is defined.

Is this script now run in parallel even outside of the downstream CI?

pearzt: > Is this script now run in parallel even outside of the downstream CI?

Not as far as I know.

> The CI_CONCURRENT_ID env var is set to timeStamp when executing outside of the Gitlab runner, so it always is defined.

True, though my comment is more about keeping the concerns between testing scripts and CI stuff separate.

silas-martens: Thank you for the quick feedback. 

I made a change to use `mktemp` to create a unique run directory where all generated files are placed. What do you think about this solution?

pearzt: LGTM

pearzt (approved):

tools/cgcollector2/test/testBase.sh
silas-martens: > The `CI_CONCURRENT_ID` env var is set to `timeStamp` when executing outside of the Gitlab runner, so it always is defined.
> 
> Is this script now run in parallel even outside of the downstream CI?

Currently not. I don't think its worth the effort.

PR #115, opened 2026-02-05
[CGFCOLLECTOR] Adding Fortran call graph collector

gomfol12:
This adds a call graph collector for Fortran. The collector is implemented as a flang plugin and generates a call graph from source-level. See the tests for a list of language features this plugin covers.

This pull request is quite large, but it is not really possible splitting it in multiply requests without pushing broken code. The tests could be separate into a different pull request, but they are tightly coupled with the collector and would fail without it, so I think keeping it in one pull request makes sense.

What this adds:
- new directory `cgfcollector` that includes flang plugin, tests, tools, etc.
- flang plugin that generates call graph in metacg format.
- two wrapper script `cgfcollector_wrapper.sh` and `cgfcollector_comp_wrapper.sh` that simplify running the plugin.
- cgCompare tool: for comparing two call graphs (can be replaced with cgdiff)
- another test runner

How to run: see `cgfcollector/README.md`

The commit history is a mess, and I don't know how strongly you value a clean git history and if I should clean this up.


cmake/FlangLLVM.cmake
```diff
@@ -0,0 +1,36 @@
...
+
+  target_include_directories(${target} SYSTEM PUBLIC ${FLANG_INCLUDE_DIRS})
+
+  find_library(
+    FLANG_LIB flang
+    PATHS ${LLVM_LIBRARY_DIR}
+    NO_DEFAULT_PATH
+  )
+
+  # target_link_libraries(${target} PUBLIC flangFrontendTool)
```
pearzt: Should this be commented out?

cgfcollector/README.md
```diff
@@ -0,0 +1,70 @@
+# CG Fortran collector
+
+## Usage
+
+The plugin is compiled into a dynamic library and can be run with the
+Flang compiler like so:
+
+`flang -fc1 -load "build/cgfcollector/libfcollector.so" -plugin "genCG"`
```
pearzt: The collector should be installed (add install rules, if necessary), so that users don't need to use the library from the build directory/

cgfcollector/CMakeLists.txt
```diff
@@ -0,0 +1,140 @@
...
+  GLOB
+  FCOLLECTOR_SOURCES
+  src/*.cpp
+)
+
+add_library(${PROJECT_NAME} SHARED ${FCOLLECTOR_SOURCES})
+add_flang(${PROJECT_NAME})
+add_metacg(${PROJECT_NAME})
+add_spdlog_libraries(${PROJECT_NAME})
+# add_json(${PROJECT_NAME})
```
pearzt: Leftover?

cgfcollector/tools/cgCompare.cpp
```diff
@@ -0,0 +1,80 @@
...
+    if (!cg2->existsAnyEdge(name1, name2)) {
+      errConsole->error("Edge from {} to {} is missing in the other call graph.", name1, name2);
+      equal = false;
+    }
+  }
+
+  return equal;
+}
+
+int main(int argc, char* argv[]) {
```
pearzt: Please have a look at CGDiff and see you can use it instead of this tool or what would be necessary to merge the tools. I would like to avoid having both tools in MetaCG.

pearzt (changes requested):
Thanks a lot for all the work, the structure looks really good. I have some initial comments, but I will try to do a more in-depth review soon.

cgfcollector/CMakeLists.txt
gomfol12: yes

cgfcollector/tools/cgCompare.cpp
gomfol12: done, it now uses CGDiff for the comparison

cmake/FlangLLVM.cmake
gomfol12: I now include it, but as I understand this, it is not really necessary. The plugin is build as a dynamic library and later loaded into the Flang binary, where the functionality of the flangFrontendTool library is already present. But I am not sure.

cgfcollector/include/Edge.h
```diff
@@ -0,0 +1,69 @@
+#pragma once
```
jplehr: Missing license header

cgfcollector/include/Edge.h
jplehr, line 1: Files should start with upper case letter.
Here and others.

cgfcollector/include/edge.h
```diff
@@ -0,0 +1,69 @@
+#pragma once
+
+#include <LoggerUtil.h>
+#include <flang/Parser/parse-tree.h>
+#include <flang/Semantics/symbol.h>
+#include <string>
+#include <vector>
+
+#include "type.h"
+#include "util.h"
```
jplehr: Project-local includes first.

cgfcollector/include/edge.h
```diff
@@ -0,0 +1,69 @@
...
+
+using namespace Fortran::semantics;
+using namespace Fortran::parser;
+using namespace metacg;
```
jplehr: No `using` in headers. That poisons potentially too many files.

cgfcollector/include/edge.h
```diff
@@ -0,0 +1,69 @@
...
+
```
jplehr: This should probably live inside the `metacg` namespace? 
If not, it should likely live in its own namespace.

What are others' opinions here?

cgfcollector/include/Edge.h
```diff
@@ -0,0 +1,69 @@
...
+#include <vector>
+
+#include "type.h"
+#include "util.h"
+
+using namespace Fortran::semantics;
+using namespace Fortran::parser;
+using namespace metacg;
+
+struct edge {
```
jplehr: Types use upper-case names.

cgfcollector/include/function.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
```
jplehr: missing license header

cgfcollector/include/ParseTreeVisitor.h
```diff
@@ -0,0 +1,244 @@
...
+    return true;
+  }
+  template <typename A>
+  void Post(const A&) {}
+
+  bool Pre(const MainProgram& p);
+  void Post(const MainProgram&);
+
+  bool Pre(const FunctionSubprogram&);
+  void Post(const FunctionSubprogram&);
```
jplehr: Why this a bit erratic whitespacing here and a bit further down?

cgfcollector/include/ParseTreeVisitor.h
```diff
@@ -0,0 +1,244 @@
...
+  std::unique_ptr<variableTracking> varTracking;
+
+  bool inFunctionOrSubroutineSubProgram = false;
+  bool inMainProgram = false;
+  bool inDerivedTypeDef = false;
+  bool inInterfaceStmt = false;
+  bool inInterfaceStmtDefinedOperator = false;
+  bool inInterfaceSpecification = false;
+
+  std::vector<edge> edges;  // added to cg in postProcess step
```
jplehr: Prefer this style: 

```suggestion
  // added to cg in postProcess step
  std::vector<edge> edges;
```

cgfcollector/src/Main.cpp
```diff
@@ -0,0 +1,149 @@
...
+    Callgraph* cg = mcgManager.getCallgraph("cg");
+    if (cg == nullptr) {
+      MCGLogger::logError("No callgraph generated");
+      return;
+    }
+
+    dot::DotGenerator dotGen(cg);
+    dotGen.generate();
+
+    std::unique_ptr<llvm::raw_pwrite_stream> dotfile = ::createOutputFile(getInstance(), getCurrentFile(), "dot");
```
jplehr: This can be problematic in parallel CI runs

cgfcollector/src/Main.cpp
```diff
@@ -0,0 +1,149 @@
...
+ *
+ */
+class CollectCGNoRename : public Fortran::frontend::PluginParseTreeAction {
+ public:
+  void executeAction() override {
+    generateCG(getParsing().parseTree(), getCurrentFile());
+
+    std::string cgString = dumpCG();
+
+    std::unique_ptr<llvm::raw_pwrite_stream> file = ::createOutputFile(getInstance(), getCurrentFile(), "");
```
jplehr: This can be problematic in parallel CI runs

cgfcollector/tools/visuel.cpp
jplehr, line 1: Should this be named `visual`?

Is this limited to the Fortran support? I would assume this can be factored out into another PR and put in as a general tool.
Other people's thoughts?

cgfcollector/CMakeLists.txt
```diff
@@ -0,0 +1,129 @@
+set(PROJECT_NAME fcollector)
```
jplehr:
```suggestion
set(PROJECT_NAME cgfcollector)
```

I think it is confusing otherwise

cmake/FlangLLVM.cmake
```diff
@@ -0,0 +1,35 @@
...
+  CONFIG
+)
+
+message(STATUS "Found FlangConfig.cmake in: ${Flang_DIR}")
+message(STATUS "Using Flang version: ${Flang_VERSION}")
+message(STATUS "Found MLIRConfig.cmake in: ${MLIR_DIR}")
+message(STATUS "Using MLIR version: ${MLIR_VERSION}")
+
+function(add_flang target)
+  target_compile_definitions(${target} PRIVATE FLANG_LITTLE_ENDIAN)
```
jplehr: Is this always the case? I tend to believe that this may be platform specific?
If so, let's put a TODO / XXX / Comment here.

jplehr (commented):
Thank you for working on this.

I did an initial brief pass, please go through the comments and address them.

One more question: Does this PR as-is already run CI pipelines? If not, please add it to the CI pipelines as appropriate.

pearzt:
Please rebase this onto the most current `devel`.

cmake/FlangLLVM.cmake
gomfol12: Is not required but leaving it in for documentation purposes

cgfcollector/include/edge.h
pearzt: We don't really have a common scheme for namespaces. Headers that are used for more than one tool (e.g., public-facing headers) should use `metacg::*`, I guess, but that's not the case here.

Until we agree on a policy, I would leave it to @gomfol12 whether to use a namespace. If we want to use a namespace, I'd suggest `metacg::` or `metacg::cgfcollector`.

cgfcollector/tools/visuel.cpp
pearzt: I agree that this is something that should extracted into a separate PR.

cgfcollector/src/Main.cpp
```diff
@@ -0,0 +1,156 @@
...
+ * @brief Like CollectCG but also generates a DOT file of callgraph
+ *
+ */
+class CollectCGwithDot : public Fortran::frontend::PluginParseTreeAction {
+ public:
+  void executeAction() override {
+    generateCG(getParsing().parseTree(), getCurrentFile());
+
+    std::string cgString = dumpCG();
+    std::unique_ptr<llvm::raw_pwrite_stream> file = ::createOutputFile(getInstance(), getCurrentFile(), "json");
```
pearzt: Please use the `MCGWriter` infrastructure for file output. For an example how to use this, have a look at `tools/cgconvert/CGConvert.cpp`.

cgfcollector/src/Main.cpp
pearzt: I think we should separate the concerns between call graph generation and visualization. cgfcollector produces the call graph. If the user want a visualization, they can use your dot export tool for this (which is a nice addition, btw).

pearzt (changes requested):

cgfcollector/src/Main.cpp
jplehr: That would indeed be nice: have the dot generator tool as a separate PR.

cgfcollector/CMakeLists.txt
```diff
@@ -0,0 +1,129 @@
...
+)
+
+configure_package_config_file(
+  ${METACG_Directory}/cmake/Config.cmake.in ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}Config.cmake
+  INSTALL_DESTINATION lib/cmake
+)
+
+install(FILES ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}Config.cmake DESTINATION lib/cmake)
+
+# visuel program to generate dot file from graph
```
jplehr:
```suggestion
# program to generate dot file from graph
```

cgfcollector/CMakeLists.txt
```diff
@@ -0,0 +1,129 @@
...
+add_executable(${PROJECT_NAME}-visuel ${CMAKE_CURRENT_SOURCE_DIR}/tools/visuel.cpp)
```
jplehr:
```suggestion
add_executable(${PROJECT_NAME}-visual ${CMAKE_CURRENT_SOURCE_DIR}/tools/visuel.cpp)
```

I think that's a typo

cgfcollector/CMakeLists.txt
```diff
@@ -0,0 +1,129 @@
...
+  configure_file_and_install
+  TEMPLATE
+  OUTPUT_BASENAME
+  DESTINATION
+)
+  set(BUILD_OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/${OUTPUT_BASENAME}")
+  set(INSTALL_OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/${OUTPUT_BASENAME}.install")
+
+  # build-time values for development
+  set(CGFCOLLECTOR_FILE_NAME "${CMAKE_CURRENT_BINARY_DIR}/lib${PROJECT_NAME}.so")
```
jplehr: Should this say that it is a library?

CGFCOLLECTOR_FILE_NAME_SO or _LIB or so?

cgfcollector/CMakeLists.txt
```diff
@@ -0,0 +1,129 @@
...
+  set(CGFCOLLECTOR_CGDIFF "${CMAKE_BINARY_DIR}/tools/cgdiff/cgdiff")
```
jplehr: Similarly should this be _BIN to indicate that's a binary?

cgfcollector/CMakeLists.txt
```diff
@@ -0,0 +1,129 @@
...
+  "test_runner.sh"
+  ""
+)
+
+configure_file_and_install(
+  "${CMAKE_CURRENT_SOURCE_DIR}/test/multi/cmake_base.txt.in"
+  "cmake_base.txt"
+  ""
+)
+file(
```
jplehr: What is this used for?

cgfcollector/CMakeLists.txt
```diff
@@ -0,0 +1,129 @@
...
+  "${CMAKE_CURRENT_BINARY_DIR}/cmake_base.txt"
+  "${CMAKE_CURRENT_SOURCE_DIR}/test/multi/cmake_base.txt"
+)
+
+configure_file_and_install(
+  "${CMAKE_CURRENT_SOURCE_DIR}/test/multi/make_base.in"
+  "make_base"
+  ""
+)
+file(
```
jplehr: What is this used for?

cgfcollector/include/Edge.h
```diff
@@ -0,0 +1,74 @@
...
+    return caller < other.caller || (caller == other.caller && callee < other.callee);
+  }
+};
+
+struct EdgeManager {
+  std::vector<Edge>& edges;
+
+  EdgeManager(std::vector<Edge>& edges) : edges(edges) {}
+
+  void addEdge(const EdgeSymbol& e, bool debug = true);
```
jplehr: What does debug indicate here and should the default really be "true"?

If that is more a development thing, the default should be false and only set to true when in debug build or when set to verbose or so.

cgfcollector/include/Procedure.h
```diff
@@ -0,0 +1,29 @@
...
+ * File: Function.h
+ * License: Part of the MetaCG project. Licensed under BSD 3 clause license. See LICENSE.txt file at
+ * https://github.com/tudasc/metacg/LICENSE.txt
+ */
+
+#pragma once
+
+#include <flang/Semantics/symbol.h>
+#include <vector>
+
```
jplehr, line 12: I think we should have this whole thing in a namespace. 

I don't remember what we did with cgcollector2, but how about metacg::cgf or so?

cgfcollector/include/ParseTreeVisitor.h
```diff
@@ -0,0 +1,252 @@
...
+
+  /**
+   * @brief Adds edges and potential finalizers edges to cg.
+   */
+  void postProcess();
+
+  // visitor methods
+
+  template <typename A>
+  bool Pre(const A&) {
```
jplehr, line 111: Why are all of those pre and post uppercase?

cgfcollector/include/ParseTreeVisitor.h
```diff
@@ -0,0 +1,252 @@
...
+  bool inMainProgram = false;
+  bool inDerivedTypeDef = false;
+  bool inInterfaceStmt = false;
+  bool inInterfaceStmtDefinedOperator = false;
+  bool inInterfaceSpecification = false;
+
+  // added to cg in postProcess step
+  std::vector<Edge> edges;
+
+  // all functions
```
jplehr: This comment does not say anything

cgfcollector/include/ParseTreeVisitor.h
```diff
@@ -0,0 +1,252 @@
...
+  std::vector<Function> functions;
+
+  // intended as a stack. It holds the current function symbol and its dummy
+  // args when the AST walker is in the respective function.
+  std::vector<Function> currentFunctions;
+
+  // all types
```
jplehr: This comment does not say anything

jplehr (commented):
Another round of comments

cgfcollector/include/edge.h
gomfol12: I have decided to put it in `metacg::cgfcollector` namespace.

cgfcollector/include/ParseTreeVisitor.h
gomfol12: This was intended to show that these visitor methods are tightly coupled. But I realized this is probably confusing for people other than me. So I removed it.

cmake/FlangLLVM.cmake
gomfol12: I am not sure anymore why I included it in the first place. I thought I had it from the example from the LLVM project, but I could not find it anymore. I think this is not necessary as Flang tries to detect the endianness anyways and set this define accordingly.

cgfcollector/tools/visuel.cpp
gomfol12: #119

cgfcollector/CMakeLists.txt
gomfol12: I opted for _LIB

cgfcollector/CMakeLists.txt
gomfol12: yes

cgfcollector/CMakeLists.txt
gomfol12: Mostly to avoid hard-coding paths to `cgfcollector_wrapper.sh` and `cgmerge2` in the tests. This tests generating a CG from a multi-file project managed by CMake.

cgfcollector/CMakeLists.txt
gomfol12: Same as `cmake_base.txt` but for make.

cgfcollector/include/ParseTreeVisitor.h
gomfol12: Because `Fortran::parser::Walk` expects them to be uppercase. This function takes a class that implements the visitor methods named in this fashion.

cgfcollector/src/Main.cpp
gomfol12: I don't understand why this could be problematic. Every test case operates on its own set of files and is only run once, per CI run.

cgfcollector/include/Edge.h
```diff
@@ -0,0 +1,82 @@
+/**
+ * File: Edge.h
+ * License: Part of the MetaCG project. Licensed under BSD 3 clause license. See LICENSE.txt file at
+ * https://github.com/tudasc/metacg/LICENSE.txt
+ */
+
+#pragma once
```
TimHeldmann: We usually work with include guards instead of #pragma once.
Should we try to keep this to one paradigm or do we not care?

cgfcollector/src/Main.cpp
```diff
@@ -0,0 +1,114 @@
...
+class CollectCG : public Fortran::frontend::PluginParseTreeAction {
+ public:
+  void executeAction() override {
+    if (Verbose) {
+      metacg::MCGLogger::instance().getConsole()->set_level(spdlog::level::debug);
+      metacg::MCGLogger::instance().getConsole()->set_pattern("%v");
+    }
+
+    // create and register callgraph
+    mcgManager.addToManagedGraphs("cg", std::make_unique<metacg::Callgraph>(), true);
```
TimHeldmann: Make name choosable via env or parameter?

cgfcollector/src/ParseTreeVisitor.cpp
```diff
@@ -0,0 +1,738 @@
...
+    const CgNode& callerNode = cg->getOrInsertNode(edge.caller);
+    const CgNode& calleeNode = cg->getOrInsertNode(edge.callee);
+
+    cg->addEdge(callerNode, calleeNode);
+  }
+}
+
+// Visitor implementations
+
+bool ParseTreeVisitor::Pre(const MainProgram& p) {
```
TimHeldmann, line 124: Add a comment, that this is not comparable to a c/c++ module, but is explicitly the "main" routine of the whole program?

cgfcollector/src/FortranUtil.cpp
```diff
@@ -0,0 +1,291 @@
...
+  // Legacy Fortran - C interoperability before BIND(C) existed.
+  // We have to do this manually because normally it would run as
+  // a pass (ExternalNameConversionPass).
+  //
+  // NOTE: underscoring can be disabled with `-fno-underscoring`
+  auto result = fir::NameUniquer::deconstruct(mangledName);
+  if (fir::NameUniquer::isExternalFacingUniquedName(result)) {
+    if (result.first == fir::NameUniquer::NameKind::COMMON && result.second.name.empty())
+      mangledName = blankCommonObjectName;
+    mangledName = GetExternalAssemblyName(result.second.name, true);
```
TimHeldmann: Query if no-underscoring flag is set and underscore accordingly

cgfcollector/src/ParseTreeVisitor.cpp
```diff
@@ -0,0 +1,738 @@
...
+  const Symbol* currentFunctionSymbol = currentFunctions.back().symbol;
+
+  // if just the name is called. (as subroutine with call and as function without call)
+  if (const Name* name = std::get_if<Name>(&p.u)) {
+    if (!name->symbol)
+      return;
+
+    // ignore intrinsic functions
+    if (name->symbol->attrs().test(Attr::INTRINSIC))
+      return;
```
TimHeldmann, line 248: Have flag to have intrinsics as part of the callgraph?
write for example might be of interes?

cgfcollector/src/ParseTreeVisitor.cpp
```diff
@@ -0,0 +1,738 @@
...
+    currentType.extendsFrom = extends.v.symbol;
+
+    MCGLogger::logDebug("Extends from: {} ({})", currentType.extendsFrom->name(), fmt::ptr(currentType.extendsFrom));
+  }
+}
+
+void ParseTreeVisitor::Post(const TypeBoundProcedureStmt& s) {
+  if (!inDerivedTypeDef)
+    return;
+
```
TimHeldmann, line 492: Give small explanation why this is relevant -> abstract types
What is the interface we are talking about?

cgfcollector/src/FortranUtil.cpp
```diff
@@ -0,0 +1,291 @@
...
+  const Fortran::semantics::DerivedTypeSpec* derived = type->AsDerived();
+  if (!derived)
+    return nullptr;
+  const Symbol* typeSymbol = &derived->typeSymbol();
+  if (!typeSymbol)
+    return nullptr;
+  return typeSymbol;
+}
+
+std::vector<const Type*> findTypeWithDerivedTypes(const std::vector<Type>& types, const Symbol* symbol) {
```
TimHeldmann: Why is this a vector?
This could be a map of "Type -> List[Implementations]"

cgfcollector/src/ParseTreeVisitor.cpp
```diff
@@ -0,0 +1,738 @@
...
+    if (!arg->hasBeenInitialized)
+      continue;
+
+    edgeM->addEdgesForFinalizers(pf);
+  }
+
+  // sort unique edges
+  std::sort(edges.begin(), edges.end());
+  auto it = std::unique(edges.begin(), edges.end());
+  edges.erase(it, edges.end());
```
TimHeldmann: Move this to the edge-manager?

cgfcollector/src/Main.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+    // determine output file name. Honors `-o` option.
+    std::string outputFile = getInstance().getFrontendOpts().outputFile;
+
+    if (outputFile.empty()) {
+      outputFile = getCurrentFile().str();
+    }
+
+    if (!NoRename) {
+      replaceExtension(outputFile, ".json");
```
TimHeldmann: We use "*.mcg" as file ending for the call-graph

cgfcollector/src/Main.cpp
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    // traverse parse tree and generate callgraph
+    std::string currentFile = getCurrentFile().str();
+
+    ParseTreeVisitor visitor(cg, currentFile);
+    Fortran::parser::Walk(getParsing().parseTree(), visitor);
+    visitor.postProcess();
+
+    mcgManager.mergeIntoActiveGraph(metacg::MergeByName());
```
TimHeldmann: Why do we use merge into active?

TimHeldmann (changes requested):

cgfcollector/include/Edge.h
gomfol12: CGDiff and pymetacg also use #pragma once. So this paradigm is already broken.

cgfcollector/src/ParseTreeVisitor.cpp
gomfol12: I added an option for this. But write is not an intrinsic. It is a statement like return. Maybe we could include a select few of them?

TimHeldmann: If open, write, and read are statements, then we should not collect them as function nodes.
We could consider having Fortran-I/O-Metadata to track usages of these functions.
This should be part of a different PR.

pearzt (approved):

PR #116, opened 2026-02-14
[CI] Update docker-run GitHub action

jplehr:
The original action seems unmaintained and results in a version error because of a too-old base image regarding docker.

This updates the action to use to a fork of the original one which updates the docker base. In addition, it pins the version to a particular commit to not just bring-in whatever changes.


pearzt (commented):
Thanks for tackling this. It seems to be the reason for all the CI failures recently. I find it somewhat weird that there is no builtin for this in GitHub CI, but apparently this is the way to do it. We could use a YAML anchor to avoid all the duplication in the file (given that's supported by GitHub?).

Also, I don't really understand why the CI is still failing for this PR.

jplehr:
> Also, I don't really understand why the CI is still failing for this PR.

I would assume I simply missed one occurrence or was running from a slightly outdated `devel` branch. I think this was done from my fork, which might have been out-of-date with upstream. I'll check+update tonight.

.github/workflows/mcg-ci.yml
jplehr: This is the missed instance.

pearzt (approved):

PR #119, opened 2026-02-24
[CGToDot] tool extracted from Fortran collector (#115)

gomfol12:
Hi,
this adds a simple tool that use the DOT generator capabilities in MetaCG to generate a DOT file from a given call graph file. I primarily used it for debugging the Fortran collector. As discussed in #115 this now lives in its own PR.

Also, not sure if this should be named `visual`. Maybe someone has a better idea about the name?


pearzt:
How about `mcg2dot` for the name of the tool? (`cg2dot` would be closer to our current naming scheme, but [tool with this name already seems to exist](https://code.google.com/archive/p/xdebugtoolkit/wikis/cg2dot.wiki))

tools/cgvisual/README.md
```diff
@@ -0,0 +1,12 @@
+# CGVisual
+
+CGVisual is a simple tool to generate a DOT file representation of a given call
+graph. It uses the existing DOT infrastructure already present in MetaCG.
```
pearzt: I think the second sentence only adds confusion. As soon as this is merged, the user doesn't care whether the infrastructure was already there before.

tools/cgvisual/visual.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include <DotIO.h>
+#include <LoggerUtil.h>
+#include <fstream>
+#include <io/MCGReader.h>
+
+static auto console = metacg::MCGLogger::instance().getConsole();
+static auto errConsole = metacg::MCGLogger::instance().getErrConsole();
+
+void printUsage() {
+  std::cout << "Usage: visuel <cg>\n\n";
```
pearzt: Note the old name `visuel` here

tools/cgvisual/visual.cpp
```diff
@@ -0,0 +1,85 @@
...
+    return EXIT_FAILURE;
+  }
+
+  std::string inputFile;
+  std::string outputFile = "callgraph.dot";
+
+  for (int i = 1; i < argc; i++) {
+    std::string arg = argv[i];
+
+    if (arg == "-h" || arg == "--help") {
```
pearzt: If you foresee more CLI options being added here in the future, I recommend using cxxopts for argument parsing since it's already a MetaCG dependency.

pearzt (changes requested):
Thanks for extracting this into a new PR. I have only minor comments. If possible, I think it would be nice to add a few simple tests to make sure we don't break this in the future.

tools/cgvisual/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+set(PROJECT_NAME CGVisual)
```
jplehr: Maybe CGVisualizer instead?
No strong feelings.

tools/cgvisual/visual.cpp
```diff
@@ -0,0 +1,85 @@
+/**
+ * File: visual.cpp
+ * License: Part of the MetaCG project. Licensed under BSD 3 clause license. See LICENSE.txt file at
+ * https://github.com/tudasc/metacg/LICENSE.txt
+ */
+
+#include <Callgraph.h>
```
jplehr: Project includes typically in the form of 

```suggestion
#include "Callgraph.h"
```

tools/cgvisual/visual.cpp
```diff
@@ -0,0 +1,85 @@
...
+    }
+  }
+
+  if (inputFile.empty()) {
+    errConsole->error("No input call graph file specified.");
+    printUsage();
+    return EXIT_FAILURE;
+  }
+
+  metacg::io::FileSource fs1(inputFile);
```
jplehr: Why `1`?

tools/cgvisual/visual.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+  auto mcgReader1 = metacg::io::createReader(fs1);
```
jplehr: Why `1`?

jplehr (commented):
Thanks for splitting this into a separate PR.
Also just minor comments.

tools/cgvisual/visual.cpp
gomfol12: Ups

tools/cgvisual/visual.cpp
gomfol12: Yes this properly makes sense. Already updated it.

tools/cgvisual/visual.cpp
gomfol12: copy-paste error

tools/cgvisual/visual.cpp
gomfol12: copy-paste error

tools/cgvisual/visual.cpp
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  if (!result.count("input")) {
+    errConsole->error("No input call graph file specified.");
+    std::cout << options.help() << std::endl;
+    return EXIT_FAILURE;
+  }
+
+  std::string inputFile = result["input"].as<std::string>();
+  std::string outputFile = result["output"].as<std::string>();
```
jplehr: Not blocking: Does it make sense to construct/use `std::filesystem` objects here?

jplehr (commented):
Only missing piece IMHO is a couple of simple tests.

tools/cgvisual/visual.cpp
gomfol12: This now makes use of std::filesystem::path and checks if the input file exists. Also shows a friendlier error message if not than just crashing.

pearzt (approved):

PR #129, opened 2026-03-11
Prepare CaGe to be extensible via plugins

TimHeldmann:
Similar to CGC2, we want users to be able to add custom collection and metadata logic to CaGe without modifying CaGe itself. 

This PR prepares CaGe to be used with plugins.

It clears up the naming confunsion, by renaming CaGe's main file from Plugin, to CaGe.
This allows the filename Plugin to now refer to CaGe plugins.
We allow to add Plugins via commandline and load them at runtime.
We augment the existing CallGraphConsumer interface to add an option CallGraphAugmentation step.


tools/cage/include/cage/generator/FileExporter.h
```diff
@@ -6,15 +6,13 @@
 #ifndef METACG_FILEEXPORTER_H
 #define METACG_FILEEXPORTER_H
 
-#include "cage/generator/CallGraphConsumer.h"
-
-#include <string>
```
jplehr, line 11: Why remove the include? 
I personally would prefer if files include what is needed and not rely on it being included somewhere else.

tools/cage/include/cage/CaGePlugin.h
```diff
@@ -0,0 +1,35 @@
...
+namespace llvm {
+class Module;
+}
+
+namespace cage {
+
+struct Plugin {
+  virtual ~Plugin() = default;
+  virtual void augmentCallGraph(llvm::Module&, metacg::Callgraph&) {}
+  virtual void consumeCallGraph(metacg::Callgraph&) {};
```
jplehr: Can we get minimum documentation here, please?
What's the intended use case for these API functions?

tools/cage/src/CaGe.cpp
```diff
@@ -30,7 +33,33 @@ static opt<cage::PTAType> pta(
...
+static list<std::string> pluginPaths("plugin-paths", desc("option list"), cat(cageOpts), CommaSeparated);
+
+
+
+
 namespace cage {
+
+Plugin* loadPlugin(const std::string& pluginPath) {
+  metacg::MCGLogger::instance().getConsole()->debug("Loading plugin");
+  void* handle = dlopen(pluginPath.c_str(), 1);
```
jplehr: I believe we rely on LLVM anyway here, right?
Can you double check whether LLVM offers helpers to work with `dlfcn`? I vaguely recall there to be helper that we use(d) in libomptarget.

tools/cage/src/CaGe.cpp
```diff
@@ -30,7 +33,33 @@ static opt<cage::PTAType> pta(
 
 static opt<std::string> cgout("cg-file", desc("Output file for the generated call graph"), cat(cageOpts), init(""));
 
+static list<std::string> pluginPaths("plugin-paths", desc("option list"), cat(cageOpts), CommaSeparated);
+
+
```
jplehr:
```suggestion
```

jplehr (changes requested):
A few comments and please reformat with clang-format.
I don't think this complies with formatting rules.

tools/cage/include/cage/CaGePlugin.h
TimHeldmann: Added documentation

tools/cage/src/CaGe.cpp
TimHeldmann: Didn't know this existed, thank you.
Swapped for LLVM's helper functions.

This seems to work based on preliminary testing, but I am not sure if they have behave identically.
If I understood correctly, one can define a search order for which LLVM's helper function traverse possible candidates for getting symbols from libraries, which I have not yet found for the dlopen/dlsym version.

tools/cage/include/cage/generator/FileExporter.h
TimHeldmann: Readded <string> include

tools/cage/src/CaGe.cpp
TimHeldmann: Removed

jplehr (commented):
Seems reasonable.
Do we have tests for it?

tools/cage/src/CaGe.cpp
```diff
@@ -30,7 +31,32 @@ static opt<cage::PTAType> pta(
...
+  }
+  metacg::MCGLogger::instance().getConsole()->trace("Getting collection object from plugin {}", pluginPath);
+  void* sym = lib.getAddressOfSymbol("getPlugin");
+  if (!sym) {
+    metacg::MCGLogger::instance().getErrConsole()->error(
+        "Could not load collectors from plugin, no Function \"getPlugin()\"!");
+    return nullptr;
+  }
+  auto getPlugin = reinterpret_cast<Plugin* (*)()>(sym);
+  Plugin* loadedPlugin=getPlugin();
```
jplehr: I think this is still not correctly formatted.

TimHeldmann: As far as I can tell, it is. 
Does it format differently on your machine?

jplehr: I would expect it to be 
```suggestion
  Plugin* loadedPlugin = getPlugin();
```

TimHeldmann: Yes you are right, I don't know why I didn't get the formatter to do it for me the first time.
Changed

TimHeldmann:
@jplehr 
> Seems reasonable. Do we have tests for it?

We currently do not have tests as part of this Repository, but @ahueck and I have developed a plugin based on this here:  https://github.com/TimHeldmann/ACGPlugin/tree/master.

I want to touch up the CGC2 plugin interface and its accompanying test plugin with the learnings gathered from this PR anyway, so I propose I add the example plugin in the second PR?

jplehr:
> I want to touch up the CGC2 plugin interface and its accompanying test plugin with the learnings gathered from this PR anyway, so I propose I add the example plugin in the second PR?

WFM

jplehr (approved):

PR #131, opened 2026-03-11
Only reiterate failure to create metadata for each node if log level is debug or finer

TimHeldmann:
When looking in the logs of our container build, I found that

1. After logging that we can not create "SomeMetadata", we additionally log that we can not create "SomeMetadata" for each node it was attached to. While I see how this could be helpful for debugging failure to create Metadata on certain nodes, I propose to limit the logging per node to debug or trace logging granularity

2. We did not log this failure for edge-based metadata, only for node and global metadata. I unified this behavior.

### DIFF
--- a/graph/src/io/VersionFourMCGReader.cpp
+++ b/graph/src/io/VersionFourMCGReader.cpp
@@ -178,7 +178,10 @@ std::unique_ptr<metacg::Callgraph> metacg::io::VersionFourMCGReader::read() {
           if (auto md = metacg::MetaData::create<>(mdKey, mdValJ, strToNode); md) {
             cg->addEdgeMetaData({nodeData.nodeId, calleeNode->getId()}, std::move(md));
           } else {
-              errConsole->debug("Could not create edge metadata of type {} for edge {} to {}", mdKey, nodeData.nodeId, calleeNode->getId());
+            if (MetaDataFactory<>::isRegistered(mdKey)) {
+              errConsole->warn("Could not create edge metadata of type {} for edge {} to {}", mdKey, nodeData.nodeId,
+                                calleeNode->getId());
+            }
             if (failedMetadataCb) {
               (*failedMetadataCb)(nodeData.nodeId, mdKey, mdValJ);
             }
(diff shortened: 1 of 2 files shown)


TimHeldmann:
In this PR I currently query the loglevel directly via spdlog's interface.
Do we want to have wrapper around that, similar to the console wrappers?

pearzt (approved):

graph/src/io/VersionFourMCGReader.cpp
```diff
@@ -179,8 +178,13 @@ std::unique_ptr<metacg::Callgraph> metacg::io::VersionFourMCGReader::read() {
           auto& mdValJ = mdElem.value();
           if (auto md = metacg::MetaData::create<>(mdKey, mdValJ, strToNode); md) {
             cg->addEdgeMetaData({nodeData.nodeId, calleeNode->getId()}, std::move(md));
-          } else if (failedMetadataCb) {
-            (*failedMetadataCb)(nodeData.nodeId, mdKey, mdValJ);
+          } else {
+            if (spdlog::get_level() <= spdlog::level::debug) {
```
jplehr: Why is this test necessary? Shouldn't the call to `warn` determine if the message needs to be printed?

pearzt: As I understand it, Tim wants the warning to be only printed when the log level is low enough. If we would just call `warn`, it would always be emitted (as long as warnings are not disabled).

On second thought, this might create some confusion. Maybe we should just use a debug print for this, not a warning?

TimHeldmann: Debug logs are usually not put to std::err but to std::out, while warning defaults to ~~std::out~~.
I am willing to change this into either.

Only change to debug, and output to std::out
Or: move to debug, but still output to std::err

What do you prefer?

Unfortunate typo:
warning defaults to **std::err**,

jplehr: > Debug logs are usually not put to std::err but to std::out, while warning defaults to std::out.

I think one of the `std::out` should read `std::err` here, right?

In case we can create  "SomeMetadata" will we always be able to create it for each node or does that depend on something?
In case one implies the other, I think, we should treat the message per node as a debug message and print a hint in the original warning.

If it can arise that MD can be created, but may fail to be created for a node, then it's a different story and we'll need to think about that a bit more.

TimHeldmann: Changed to only emit a warning on std::err if the metadata should be emitable (aka. metadata key is registered in the metadatafactory), but could not be created successfully.

jplehr (approved):
How about changing commit title to something like this:
```
[Graph] Emit create-MD error per node/edge in debug-level
```

Can you give it one more `git-clang-format` just to be sure? Thanks!

PR #133, opened 2026-03-11
[Graph] Add Callgraph class unit tests

jplehr:
This is in an effort to improve the testing and remove the concerns about the state of our testing across the graph library.
Just adds some initial API level tests and uncovered a bug that is also fixed.


pearzt (approved):
LGTM

PR #135, opened 2026-03-18
[Graph] Add more Callgraph unit tests

jplehr:
While adding the unit tests, I also found a fixed a simple bug in the return value of one of the API functions.


jplehr:
While working on this, I realized that many of the `MCGManagerTest.cpp` tests are actually `Callgraph` tests.
In a next session I'm going to refactor them / split them out / move them into the appropriate place.

pearzt (approved):
LGTM

PR #138, opened 2026-03-27
Feat/plugin interface touchup

TimHeldmann:
This PR adds two Demoplugins, one for CaGe and one for CGC2.
They are also added to the CI, to test whether we can correctly link and run plugins in CaGe and CGC2.
To add the Demoplugins the PR adds an interface library to link against, as well as changin the CMake files to install the necessary header files.

It lastly moved the existing FileInfoMetadata+Collector combo, which served as a demo plugin for CGC2 to the graphlib core-metadata. 
This is because legacy CGC outputs FileInfoMD by default.

Looking forward to your feedback.

PR #139, opened 2026-04-29
Fix Downstream CI

TimHeldmann:
The downstream CI was red since PR #138.
This PR fixes the misconfiguration in the build directories, as well as adds compatibility with LLVM/Clang 15, as this version is part of the downstream CI's compatibility matrix.


pearzt (approved):
LGTM

PR #142, opened 2026-05-28
[Downstream CI] Fix build job limt

pearzt:
As it turns out, CMake counterintuitively only respects `CMAKE_BUILD_PARALLEL_LEVEL` when *not* passing `--parallel`.


jplehr:
So this change still makes the build run in parallel?
That's indeed counter intuitive.

pearzt:
> So this change still makes the build run in parallel?

Yes, since `CMAKE_BUILD_PARALLEL_LEVEL` is set. Passing `--parallel` makes it use Ninja's default number of processes (104, in our case), indepedent of `CMAKE_BUILD_PARALLEL_LEVEL`.

https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-build-j

jplehr (approved):

TimHeldmann (approved):

PR #143, opened 2026-06-03
[NFC, pymetacg] Use Python_SITEARCH for default installation path

pearzt:
I just stumbled over `Python_SITEARCH` which is populated by `find_package(Python)`. This makes it easier to set the default install directory for pymetacg without weird path construction we used before.

### DIFF
--- a/pymetacg/CMakeLists.txt
+++ b/pymetacg/CMakeLists.txt
@@ -18,9 +18,7 @@ nanobind_disable_stack_protector(pymetacg)
 add_metacg_library(pymetacg)
 
 if(NOT CMAKE_INSTALL_PYTHON_LIBDIR)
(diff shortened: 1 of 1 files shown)


TimHeldmann (approved):
Looks like a simplification to me, if it works, we can ship this.
Any reason to use a separate variable in the first place?
Why not use ${Python_SITEARCH} itself as the install destination?

pearzt:
> Why not use ${Python_SITEARCH} itself as the install destination?

Because there can be scenarios where you want to control where the Python stuff is installed. In my case, I set `CMAKE_INSTALL_PYTHON_LIBDIR` in my Spack package recipe for MetaCG so that Spack picks up on the Python module.

Issue #145, opened 2026-06-12
Compile error building against clang 22


PhilipDeegan:
```
/usr/include/clang/AST/TypeNodes.inc:50:1: error: too many arguments to function call, expected single argument 'TL', have 2 arguments
   50 | TYPE(FunctionProto, FunctionType)
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/clang/AST/TypeLocNodes.def:32:27: note: expanded from macro 'TYPE'
   32 | #define TYPE(Class, Base) UNQUAL_TYPELOC(Class, Base##Loc)
      |                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/clang/AST/TypeLocNodes.def:24:39: note: expanded from macro 'UNQUAL_TYPELOC'
   24 | #  define UNQUAL_TYPELOC(Class, Base) TYPELOC(Class, Base)
      |                                       ^~~~~~~~~~~~~~~~~~~~
/usr/include/clang/AST/RecursiveASTVisitor.h:736:50: note: expanded from macro 'TYPELOC'
  735 |     return getDerived().Traverse##CLASS##TypeLoc(TL.castAs<CLASS##TypeLoc>(),  \
      |            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  736 |                                                  TraverseQualifier);
      |                                                  ^~~~~~~~~~~~~~~~~
/home/deegan/git/MetaCG/cgcollector/lib/src/AliasAnalysis.cpp:664:31: note: in instantiation of member function 'clang::RecursiveASTVisitor<implementation::ASTInformationExtractor>::TraverseTypeLoc' requested here
  664 |   return RecursiveASTVisitor::TraverseTypeLoc(TL);
      |                               ^
/home/deegan/git/MetaCG/cgcollector/lib/src/AliasAnalysis.cpp:991:31: note: 'TraverseFunctionProtoTypeLoc' declared here
  991 | bool ASTInformationExtractor::TraverseFunctionProtoTypeLoc(clang::FunctionProtoTypeLoc TL) {
      |                               ^                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2 warnings and 2 errors generated.
[...]

PR #146, opened 2026-06-12
[FIX] Adds compatibility for LLVM-22 

TimHeldmann:
In LLVM-22 two of the type traversal functions got a new parameter, whether to traverse qualifiers.

We do not care about the qualifiers, so we just need to pass them along.
This PR uses macros to define the correct function and pass the qualifiers accordingly if available.

This should address #145.


pearzt (approved):
LGTM

jplehr (approved):

pearzt (changes requested):
Downstream CI still fails when building with LLVM 22. Let's investigate internally, @TimHeldmann.

pearzt (approved):
Build failures are resolved

PR #147, opened 2026-06-12
[CI] Symlink flang to flang-new if flang is not in the PATH

pearzt:
This is to fix the downstream CI test failures for cgfcollector.

CC @gomfol12


TimHeldmann (approved):
While I think, it is unfortunate to even need to do something like this,
I agree this is probably the best solution.

LGTM

PR #151, opened 2026-07-22
[CaGe] Add embedding plugin

sebastiankreutzer:
This patch adds basic functionality to embed the generated CaGe call graph into a dedicated `metacg` ELF section.
The `extract_metacg` script is added in `utils` to facilitate extracting this data.
Tests are amended to check that the embedded graph matches the one written to a normal file.

Follow-up PRs will extend this functionality to introduce new merge policies based on linkage metadata to enable correct merging of shared libraries.

### DIFF
--- a/tools/cage/src/CaGe.cpp
+++ b/tools/cage/src/CaGe.cpp
@@ -6,6 +6,7 @@
 #include "cage/CaGe.h"
 #include "cage/interface/CaGePlugin.h"
 
+#include "cage/generator/CallGraphEmbedder.h"
 #include "cage/generator/CallgraphGenerator.h"
 #include "cage/generator/FileExporter.h"
 
@@ -26,6 +27,8 @@ static opt<cage::PTAType> pta(
     cat(cageOpts), init(cage::PTAType::No));
 
 static opt<std::string> cgout("cg-file", desc("Output file for the generated call graph"), cat(cageOpts), init(""));
+
+static opt<bool> embed("embed", desc("Embed output graph into binary"), cat(cageOpts), init(true));
 
 static list<std::string> pluginPaths("plugin-paths", desc("option list"), cat(cageOpts), CommaSeparated);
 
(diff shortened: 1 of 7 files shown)


tools/cage/src/CaGe.cpp
```diff
@@ -72,6 +75,9 @@ PreservedAnalyses CaGe::run(Module& M, ModuleAnalysisManager& MA) {
 
   Generator gen(pta);
   gen.addPlugin(std::make_unique<FileExporter>(outfile));
+  if (embed) {
+    gen.addPlugin(std::make_unique<GraphEmbedder>(M));
+  }
```
TimHeldmann: Now that we have more than one way to store the callgraph we should probably make these configurable.
If I understand correctly we always output the CG, but only if embed is set, we put it in the binary.
A user might want to only embed and not output the cg.
Maybe we should change this to an outputtype enum, with file and embed, kinda like this:

```
enum class OutputTypes {
  File,
  Embed
};
static bits<OutputTypes> outputTypeBits(
    desc("Where to output the graph"),
    values(clEnumValN(OutputTypes::File, "File", " Output into separate file"),
           clEnumValN(OutputTypes::Embed, "Embed", "Embed the graph into the metacg section of the binary"),
    cat(cageOpts));
```

utils/extract_metacg
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+objcopy -O binary --only-section=metacg "$1" /dev/stdout | strings -n 1
```
TimHeldmann, line 2: Why do we need the pipe to strings?

TimHeldmann (changes requested):
Only one nit and a question.

Thank you for this feature,

Tim

tools/cage/src/CaGe.cpp
sebastiankreutzer: Sounds reasonable. Does this allow enabling both options? This would be convenient in some cases, and needed for testing.

utils/extract_metacg
sebastiankreutzer: IIRC this was needed because the string is 0-terminated

tools/cage/src/CaGe.cpp
TimHeldmann: Yes, as far as I know it allows for a list. You can steal the implementation from cgc2's collector list.

